PR #3, opened 2019-09-17
[EPMLSTRCMW-97] Git hooks life cycle scripts

DenisVerkhoturov:
Initialize a `git-hooks` lifecycle so we can install it once and it will be updated by itself if needed. With hooks, we can automate a bunch of verification work, such as validating that commit messages meet the guideline.


AlekseiLitkovetc:
@DenisVerkhoturov could we please create a task for this?
Just to have an ability to track it later.

likeanowl:
@DenisVerkhoturov Great request!
I think that addition of some documentation would be extremely helpful for newcomers.

PR #4, opened 2019-09-23
[EPMLSTRCMW-51] Create User DAO

AlekseiLitkovetc:

PR #6, opened 2019-09-27
chore: Add .scalafmt.conf

likeanowl:
This config file has come from a perfect world we will never see...
(mixed from zio/zio and circe/circe)
Also, two sbt commands were added, `fmt` and `check`.

Resolves: #EPMLSTRCMW-96

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,21 @@
+version = "2.0.1"
+maxColumn = 120
+align = some
+continuationIndent.defnSite = 2
+docstrings = JavaDoc
+assumeStandardLibraryStripMargin = true
+lineEndings = preserve
+newlines.alwaysBeforeElseAfterCurlyIf = false
+newlines.alwaysBeforeMultilineDef = false
+optIn.breakChainOnFirstMethodDot = false
+spaces.inImportCurlyBraces = true
+includeCurlyBraceInSelectChains = false
+danglingParentheses = true
+optIn.annotationNewlines = true
+rewrite.rules = [
+  AvoidInfix,
+  RedundantBraces,
+  RedundantParens,
+  AsciiSortImports,
+  PreferCurlyFors
+]
--- a/build.sbt
+++ b/build.sbt
@@ -8,21 +8,46 @@ ThisBuild / scalacOptions := Seq(
     "-Xfatal-warnings"
 )
 
+lazy val formatAll = taskKey[Unit]("Format all the source code which includes src, test, and build files")
+lazy val checkFormat = taskKey[Unit]("Check all the source code which includes src, test, and build files")
+
+lazy val commonSettings = Seq(
+  formatAll := {
+    (scalafmt in Compile).value
+    (scalafmt in Test).value
+  },
+  checkFormat := {
+    (scalafmtCheck in Compile).value
+    (scalafmtCheck in Test).value
+  },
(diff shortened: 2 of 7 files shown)


.scalafmt.conf
```diff
@@ -0,0 +1,23 @@
+version = "2.0.1"
+maxColumn = 120
+align = most
```
likeanowl: This one is actually kinda controversal. E.g. it could generate smth like this:
`val instance: Config              = ConfigFactory.load(commonConfigPath).getConfig(commonConfigNamespace)`
Should we keep it? Or mb just set it to `none`? I want to discuss this.

build.sbt
```diff
@@ -1,19 +1,19 @@
 import Dependencies._
 
-lazy val defaultSettings = Seq(
```
myazinn: @AlekseiLitkovetc made some changes on the structure of the build.sbt in his [PR](https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/pull/4/files). Maybe we should not change the formatting of this file concurrently? :)

likeanowl: I could take care about merging it, don't worry :)

.scalafmt.conf
```diff
@@ -0,0 +1,23 @@
...
+continuationIndent.defnSite = 2
+docstrings = JavaDoc
+assumeStandardLibraryStripMargin = true
+docstrings = JavaDoc
```
ilyaonishenko: Looks like repetition (line 5) for me

build.sbt
```diff
@@ -1,19 +1,19 @@
...
-  defaultScalacOptions)
+lazy val defaultSettings =
+  Seq(name := "cromwell_pipeline", version := "0.1", scalaVersion := "2.12.9", defaultScalacOptions)
 
-lazy val cromwell_pipeline = (project in file("."))
-    .settings(
-      defaultSettings, 
-      libraryDependencies ++= akkaDependencies ++ dbDependencies ++ testDependencies
-    )
+lazy val cromwell_pipeline = (project in file(".")).settings(
```
ilyaonishenko: If it is the result of formatting, personally I prefer to have `.settings` on the next line

.scalafmt.conf
ilyaonishenko: I would propose to remove it

.scalafmt.conf
likeanowl: Yes, should've double-checked. Will fix

.scalafmt.conf
likeanowl: I think that the best option is `more`, bc it will align only pattern-matching arrows

.scalafmt.conf
```diff
@@ -0,0 +1,23 @@
+version = "2.0.1"
+maxColumn = 120
+align = most
+continuationIndent.defnSite = 2
+docstrings = JavaDoc
```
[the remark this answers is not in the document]
likeanowl: It's just looks nicer IMHO. The only difference between `JavaDoc` and `ScalaDoc` is by which asterisk the comment would be aligned. For `JavaDoc` it is 1st and for `ScalaDoc` it is 2nd

likeanowl:
I've made some changes in build.sbt and also rebased this branch against latest `develop` and applied reconfigured formatter.

likeanowl:
I've also rebased this branch against latest `develop`  and resolved merge conflicts. Please take a look

PR #8, opened 2019-10-14
EPMLSTRCMW-102: start db in container

kynyan:
- added docker-compose.yml

### DIFF
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3"
+services:
+  #  Create a service named db.
+  db:
+    image: "postgres:12-alpine"
+    container_name: "pipeline_db"
+    #   Setup the username, password, and database name
+    environment:
+      - POSTGRES_USER=pipeline
+      - POSTGRES_PASSWORD=pipeline
+      - POSTGRES_DB=pipeline
+    #   Map port 6432 (localhost) to port 5432 on the container
+    ports:
+      - "6432:5432"


scripts/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: "postgres:12-alpine"
+    container_name: "pipeline_db"
+    #   Setup the username, password, and database name
+    environment:
+      - POSTGRES_USER=pipeline
+      - POSTGRES_PASSWORD=pipeline
+      - POSTGRES_DB=pipeline
+    #   Map port 6432 (localhost) to port 5432 on the container
+    ports:
+      - "6432:5432"
```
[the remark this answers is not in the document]
kynyan: Yes, this is to avoid conflicts with locally running Postgres

scripts/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  #  Create a service named db.
+  db:
+    image: "postgres:12-alpine"
+    container_name: "pipeline_db"
+    #   Setup the username, password, and database name
+    environment:
+      - POSTGRES_USER=pipeline
+      - POSTGRES_PASSWORD=pipeline
+      - POSTGRES_DB=pipeline
```
DenisVerkhoturov: I suggest we omit the actual values from the `compose`-file and use [`host-specific`](https://docs.docker.com/compose/compose-file/#environment) definitions instead.

It would be also nice to provide default `.env`-file with default values for all the required environmental variables.

scripts/docker-compose.yml
DenisVerkhoturov: Please, add a new line at the end of the file.

DenisVerkhoturov: There is no need to expose the port for the `db` service to the host machine since it is not supposed to be reachable from the outer world. Only the application service should have a connection to it over the common [`network`](https://docs.docker.com/compose/compose-file/#networks).

kynyan: I need to know more about the planned infrastructure setup, is there any doc I could refer to? How will we debug (for example, db migrations) without connecting remotely to DB? I know there are ways to connect to container from the host it's running on, but it's more complicated than just creating a tunnel to the exposed port.

kynyan: Besides, for now there will be only db in that network :)

DenisVerkhoturov:
Please, install `hooks` as it is inquired in the [`Installation`](https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/tree/develop#installation) section of the `README`-file and update the commit message to accord the [style guide](https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/blob/develop/docs/git-commit-message-style-guide.md) that is used on the project.

kynyan:
Closing this request as it appeared to have no value: it's easy to start db locally without docker-compose. We may return to docker-compose, when ready to combine db and an app (may be some other services)

PR #9, opened 2019-10-16
[EPMLSTRCMW-101] Ignore comments in git messages

SergeySdv:
Fixing of ignore comments in git messages

### DIFF
--- a/hooks/commit-msg.d/check-if-message-accords-with-style-guide.sh
+++ b/hooks/commit-msg.d/check-if-message-accords-with-style-guide.sh
@@ -6,7 +6,7 @@
 # characters length. The body is considered everything until the end of the message.
 pattern="^(feat|fix|docs|style|refactor|test|chore):[ ](.{1,50})("$'\n'"("$'\n'"[^"$'\n'"]{0,72})+)?$"
 
-message=$(cat "${1}")
+message=$(grep --invert-match "^#" < "${1}")
 
 if [[ "${message}" =~ ${pattern} ]]; then
   exit 0


hooks/commit-msg.d/check-if-message-accords-with-style-guide.sh
```diff
@@ -6,7 +6,7 @@
 # characters length. The body is considered everything until the end of the message.
 pattern="^(feat|fix|docs|style|refactor|test|chore):[ ](.{1,50})("$'\n'"("$'\n'"[^"$'\n'"]{0,72})+)?$"
 
-message=$(cat "${1}")
+message=$(grep --invert-match "^#" < "${1}")
```
likeanowl: Why not just use `-v`? I think this is too verbose.

DenisVerkhoturov: @likeanowl for documentation purposes it is better to use full syntax in scripts rather than short one.

likeanowl: I can agree with this, but first link when googling "grep invert match" will suggest a short argument$ do we really want to be so verbose?

DenisVerkhoturov: @likeanowl, you obviously will use a short syntax when you want to run a single command. But in scripts, it is better to be as verbose as you can, for the same reason we do not use single letter variables when writing in other languages - readability.

likeanowl: @DenisVerkhoturov OK.

SergeySdv:
fixed

likeanowl:
Great work! 👍 
@SergeySdv please consider removing extra "fix" from commit message.

hooks/commit-msg.d/check-if-message-accords-with-style-guide.sh
SergeySdv: photo_2019-10-17_20-33-34

likeanowl:
Good fix! 👍 
Since you have fixed commit message, could you also please fix a PR name? And when this is done, I think that I'll accept this PR.

hooks/commit-msg.d/check-if-message-accords-with-style-guide.sh
likeanowl: @andreibelov I still think that `--invert-match` is absolutely OK here, since this script is quite simple and this argument could be googled very easy.

PR #11, opened 2019-10-22
[EPMLSTRCMW-104] Create auth integration test

AlekseiLitkovetc:


AlekseiLitkovetc:
@likeanowl thank you for your review!

portal/src/it/scala/cromwell/pipeline/controller/AuthControllerItTest.scala
```diff
@@ -0,0 +1,100 @@
...
+          checkAuthTokens
+        }
+      }
+    }
+
+    "refresh" should {
+
+      "return updated token headers if refresh token was valid and active" in {
+        val newUser = getDummyUser(userPassword)
+        awaitForResult {
```
timaliberdov: You could just use [this](http://doc.scalatest.org/3.0.1-2.12/org/scalatest/concurrent/Futures$FutureConcept.html#futureValue(implicitconfig:Futures.this.PatienceConfig,implicitpos:org.scalactic.source.Position):T). Also take a look at [this](http://doc.scalatest.org/3.0.1-2.12/org/scalatest/concurrent/ScalaFutures.html#whenReady[T,U](future:Futures.this.FutureConcept[T])(fun:T=%3EU)(implicitconfig:Futures.this.PatienceConfig,implicitpos:org.scalactic.source.Position):U).

portal/src/it/scala/cromwell/pipeline/controller/AuthControllerItTest.scala
```diff
@@ -0,0 +1,100 @@
...
+import cromwell.pipeline.BaseItTest
+import cromwell.pipeline.controller.AuthController._
+import cromwell.pipeline.datastorage.dto.auth.{SignInRequest, SignUpRequest}
+import cromwell.pipeline.datastorage.dto.{User, UserId}
+import cromwell.pipeline.utils.StringUtils
+import org.scalatest.compatible.Assertion
+import play.api.libs.json.Json
+
+
+class AuthControllerItTest extends BaseItTest with ScalatestRouteTest with ForAllTestContainer {
```
kynyan: I'm a scala newbie, but from the OOP point it's not a good practice to use inheritance in tests. It'd be great to replace it with composition if possible. Here is the first link given by google (for Java) https://javarevisited.blogspot.com/2013/06/why-favor-composition-over-inheritance-java-oops-design.html

AlekseiLitkovetc: @kynyan, thank you for your proposal,

I've investigated the article on which you refer but noticed only one thesis about tests (where the mocking process was described). Here we have integration testing and we do not need to mock something.

With respect to our unit tests, we can easily mock other components and I don't see any reason to change something in current logic.

Could you please describe your suggestion in more detail what exact changes you would like to add?

kynyan: Ok, I'm not sure if I can find anything specifically about tests, may be [this one](https://www.petrikainulainen.net/programming/unit-testing/3-reasons-why-we-should-not-use-inheritance-in-our-tests/) is closer. And it's hard to give a concrete example for this project, but it all comes down to the limitations introduced by inheritance. Once you extend your test from something and then you need to change some part of what you inherit for another test, you are going to make duplicated code, and there is also a risk of breaking child's functionality. 

I've heard "the rule of thumb": use inheritance to model **is-a** relationships, and composition to model **has-a** relationships. `AuthControllerItTest` just uses (**has**) access to the DB which is encapsulated in `BaseTest`. In Java I would create annotation `@RepositoryTest` which would enable DB config for the test, but I don't know if it can be done with Scala.

ilyaonishenko: I agree with @kynyan. We should try to use composition instead of inheritance. 
At least, maybe I will need `beforeEach` to have other behaviour than defined right now in the `BaseItTest`. So I will need to override something that you've already overrode. So what is the point.

portal/src/it/scala/cromwell/pipeline/controller/AuthControllerItTest.scala
AlekseiLitkovetc: Thank you, the implementation has become easier.

portal/src/it/scala/cromwell/pipeline/controller/AuthControllerItTest.scala
AlekseiLitkovetc: @kynyan, yes, the second article is closer and I've got what you're talking about. Thank you!

You can check the updated version when you will have a few minutes.

portal/src/it/scala/cromwell/pipeline/BaseItTest.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.typesafe.config.ConfigFactory
+import org.scalatest.{BeforeAndAfterEach, Matchers, WordSpec}
+
+import scala.collection.convert.ImplicitConversions._
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, Future}
+
+trait BaseItTest extends WordSpec with Matchers with BeforeAndAfterEach {
+
+  private implicit val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
```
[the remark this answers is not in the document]
AlekseiLitkovetc: It's a good notice, thank you!

portal/src/it/scala/cromwell/pipeline/controller/AuthControllerItTest.scala
kynyan: Looks good to me, thanx!

PR #15, opened 2019-11-29
feat: add docker-compose

Benderbej:


likeanowl:
This is a great work, however I have some thoughts:
Do we really need docker-compose for just one container? Wouldn't it be better to create a single Dockerfile with container, simple port forwarding and db creation? I think that it would be way more appropriate in our case, because docker-compose is an overkill for a standalone container with db.

likeanowl:
What do you think about it? @AlekseiLitkovetc @mnesvit @ilyaonishenko @Mikhail-Sosnin

Mikhail-Sosnin:
@Benderbej well done!

@likeanowl thanks for the questions.
> This is a great work, however I have some thoughts:
> Do we really need docker-compose for just one container? 
If I clearly understand environment goals this benefits us from the local/cloud deployment perspective also this might be extended in any time. 
@andreibelov @mnesvit @timaliberdov  please correct me if I'm mistaken.
> Wouldn't it be better to create a single Dockerfile with simple port forwarding and db creation? 
And build manually the image, etc, instead of the simple 'docker-compose up'? 
And later that replace it with the compose again? 
Probably this might be a kind of the work duplication.
>I think that it would be way more appropriate in our case, because docker-compose is an overkill for a standalone container with db.
Same as above.

andreibelov:
It's probably better to have any DevOps related stuff in separate repo which is already created

https://github.com/EpamLifeSciencesTeam/cmwl_pipeline_devops

PR #20, opened 2019-12-10
feat: implement unit tests for UserRepository

srdkr:
Testcontainers functionality was added to unit tests.
Now classpath of 'test' is shared with 'it:test'
(common classes should be moved from 'src/it/' to '/src/test').
Also, postgresql test containers are now mapped to random ports.


Mikhail-Sosnin:
Well done!

PR #21, opened 2019-12-12
add Gitlab CE to docker-compose

srdkr:
UI and repos are available via http protocol on port 9080.

Resolves: #EPMLSTRCMW-112

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,3 +12,20 @@ services:
     #   Map port 6432 (localhost) to port 5432 on the container
     ports:
       - "6432:5432"
+
+  gitlab:
+    image: 'gitlab/gitlab-ce:latest'
+    restart: always
+    environment:
+      GITLAB_OMNIBUS_CONFIG: |
+       external_url 'http://localhost:9080'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -11,4 +11,22 @@ services:
...
     #   Map port 6432 (localhost) to port 5432 on the container
     ports:
-      - "6432:5432"
\ No newline at end of file
+      - "6432:5432"
+
+  gitlab:
+    image: 'gitlab/gitlab-ce:latest'
+    restart: always
+  #  hostname: 'gitlab-pipeline.cmwl'
```
ilyaonishenko: Why do we need this?

srdkr: well, that line was from an official example (https://gitlab.com/gitlab-org/omnibus-gitlab/raw/master/docker/docker-compose.yml),
so I commented out that line and left it just in case

ilyaonishenko: I don't think we will need it soon. 
Could you delete it, please?

srdkr: sure, done

PR #22, opened 2019-12-16
feat: add project entity dao

srdkr:
Resolves: #EPMLSTRCMW-71

### DIFF
--- a/portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
+++ b/portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
@@ -1,0 +1,34 @@
+package cromwell.pipeline.datastorage.dao.entry
+
+import cromwell.pipeline.datastorage.Profile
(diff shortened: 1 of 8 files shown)


portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
```diff
@@ -0,0 +1,34 @@
+package cromwell.pipeline.datastorage.dao.entry
+
+import cromwell.pipeline.datastorage.Profile
+import cromwell.pipeline.datastorage.dto.{ Project, ProjectId, UserId }
+
+trait ProjectEntry extends Profile {
+  self: UserEntry =>
```
ilyaonishenko: Looks like sleftypes are the only way indeed.
Just a small sugar request: make it `trait ProjectEntry { this: Profile with UserEntry =>`.

And the same thing with `UserEntry`

srdkr: updated both entries

PR #16, opened 2019-12-03
EPMLSTRCMW 108

YuliyaSheyko:
Add deactivation feature with email an user's id

PR #23, opened 2019-12-16
Epmlstrcmp 109

Dmitrii-Shevchenko:
Added tests, fixed notices

### DIFF
--- a/portal/src/main/scala/cromwell/pipeline/controller/UserController.scala
+++ b/portal/src/main/scala/cromwell/pipeline/controller/UserController.scala
@@ -1,0 +1,43 @@
+package cromwell.pipeline.controller
+
+import akka.http.scaladsl.server.Directives.complete
+import cromwell.pipeline.datastorage.dto.user.UserUpdateRequest
+import cromwell.pipeline.service.UserService
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import cromwell.pipeline.datastorage.dto.user.PasswordUpdateRequest
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+
+import scala.concurrent.ExecutionContext
+import scala.util.{ Failure, Success }
+
+class UserController(userService: UserService)(implicit executionContext: ExecutionContext) {
+
+  val route: Route =
+    path("users" / Segment) { id =>
+      concat(
+        put {
+          entity(as[UserUpdateRequest]) { request =>
+            onComplete(userService.updateUser(id, request)) {
+              case Success(count) => getStatus(count)
+              case _              => complete(StatusCodes.InternalServerError)
+            }
+          }
+        },
+        put {
+          entity(as[PasswordUpdateRequest]) { request =>
+            onComplete(userService.updateUserPassword(id, request)) {
+              case Success(count) => getStatus(count)
+              case Failure(_)     => complete(StatusCodes.BadRequest, "Password doesn't match")
+              case _              => complete(StatusCodes.InternalServerError)
+            }
+          }
+        }
+      )
+    }
+
+  private def getStatus(count: Int): Route =
+    if (count == 1) complete(StatusCodes.NoContent)
+    else complete(StatusCodes.BadRequest, "Specified ID doesn't exist")
+}
--- a/portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
+++ b/portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
@@ -31,4 +31,12 @@ class UserEntry(val pipelineDatabaseEngine: PipelineDatabaseEngine) {
 
   def addUserAction(user: User) = (users.returning(users.map(_.userId))) += user
 
+  def updateUser(updatedUser: User) =
+    users
+      .filter(_.userId === updatedUser.userId)
+      .map(user => (user.email, user.firstName, user.lastName))
+      .update((updatedUser.email, updatedUser.firstName, updatedUser.lastName))
+
+  def updateUserPassword(updatedUser: User) =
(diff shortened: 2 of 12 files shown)


portal/src/main/scala/cromwell/pipeline/controller/UserController.scala
```diff
@@ -0,0 +1,43 @@
...
+              case _              => complete(StatusCodes.InternalServerError)
+            }
+          }
+        }
+      )
+    }
+
+  private def getStatus(count: Int): Route =
+    if (count == 1) complete(StatusCodes.NoContent)
+    else complete(StatusCodes.BadRequest, "Specified ID doesn't exist")
```
myazinn: As we have repeatedly discussed during meetings, this logic should be in the `Service` layer.

If no user was updated during an update of the user, the `Service` should return a `Future.failed` with some meaningful error (e.g. `IllegalArugmentException`). The `Controller` should never make any decision on whether it is good or bad that exactly one user was updated. `Controller` must only provide a way for our application to communicate with the outside world.

portal/src/main/scala/cromwell/pipeline/controller/UserController.scala
```diff
@@ -0,0 +1,43 @@
...
+              case _              => complete(StatusCodes.InternalServerError)
+            }
+          }
+        },
+        put {
+          entity(as[PasswordUpdateRequest]) { request =>
+            onComplete(userService.updateUserPassword(id, request)) {
+              case Success(count) => getStatus(count)
+              case Failure(_)     => complete(StatusCodes.BadRequest, "Password doesn't match")
+              case _              => complete(StatusCodes.InternalServerError)
```
myazinn: What do you think should be the result of the `updateUserPassword` method so that we get to this case?

portal/src/main/scala/cromwell/pipeline/datastorage/dto/User.scala
```diff
@@ -1,16 +1,24 @@
...
-  email: String,
-  passwordHash: String,
-  passwordSalt: String,
-  firstName: String,
-  lastName: String,
+  email: String = "",
+  passwordHash: String = "",
+  passwordSalt: String = "",
+  firstName: String = "",
+  lastName: String = "",
```
myazinn: Using an empty string as a placeholder is a terrible idea. As @ilyaonishenko said, only bad guys do things like this.
If you are planning to use a case class, you _must_ provide a valid values for all of its fields. If you don't have these values, you should not use this case class at all. In this case, you may create your own case class, for example.

portal/src/main/scala/cromwell/pipeline/datastorage/dto/User.scala
```diff
@@ -1,16 +1,24 @@
...
   profilePicture: Option[ProfilePicture] = None
 )
 
 final case class UserId(value: String) extends MappedTo[String]
+object UserId {
+  implicit val userIdFormat: OFormat[UserId] = Json.format[UserId]
```
myazinn: `UserId` should be formatted differently. You can find how this should be done at [this](https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/pull/16) PR.

portal/src/main/scala/cromwell/pipeline/controller/UserController.scala
```diff
@@ -0,0 +1,43 @@
...
+          entity(as[UserUpdateRequest]) { request =>
+            onComplete(userService.updateUser(id, request)) {
+              case Success(count) => getStatus(count)
+              case _              => complete(StatusCodes.InternalServerError)
+            }
+          }
+        },
+        put {
+          entity(as[PasswordUpdateRequest]) { request =>
+            onComplete(userService.updateUserPassword(id, request)) {
```
myazinn: Ah, I realized that we did not notice the elephant in the room. You never check that `id` of the user to update is the same as the `id` of the current user. This means that one user can easily send a request to update any other user. This is a huge security hole.

portal/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  def updateUser(id: String, request: UserUpdateRequest): Future[Int] = {
+    val updatedUser =
+      User(UserId(id), email = request.email, firstName = request.firstName, lastName = request.lastName)
+    userRepository.updateUser(updatedUser)
+  }
+
+  def updateUserPassword(id: String, request: PasswordUpdateRequest): Future[Int] =
+    if (request.newPassword != request.repeatPassword) {
+      Future.failed(Exception(JsError("Password doesn't match")))
```
myazinn: Do not use `json` in the `Service` layer. The existence of the `json` in our application is an implementation detail of the `Controllers`, `Service` layer should not know about it.

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
```diff
@@ -31,4 +31,12 @@ class UserEntry(val pipelineDatabaseEngine: PipelineDatabaseEngine) {
...
   def addUserAction(user: User) = (users.returning(users.map(_.userId))) += user
 
+  def updateUser(updatedUser: User) =
+    users
+      .filter(_.userId === updatedUser.userId)
+      .map(user => (user.email, user.firstName, user.lastName))
+      .update((updatedUser.email, updatedUser.firstName, updatedUser.lastName))
+
+  def updateUserPassword(updatedUser: User) =
+    users.filter(_.userId === updatedUser.userId).map(user => (user.passwordHash)).update((updatedUser.passwordHash))
```
ilyaonishenko: You don't have to work with `User` entity here.
You're working with `UserUpdateRequest` and that's great! So pass it here.
If you do this, you won't be forced to redefine `User` entity with empty lines or do other not-nice stuff.

Dmitrii-Shevchenko:
https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/pull/29

PR #27, opened 2019-12-30
Cloud migration

Mikhail-Sosnin:


iliaonishchenko (approved):

andreibelov (approved):
No objections from my side

DenisVerkhoturov (changes requested):
Thank you for the pull-request!

You can find minor observations below. In addition to them, could you, please, format the stages in the Dockerfile consistently so that for both stages it would either use empty lines between commands or not.

PR #30, opened 2020-01-16
feat: gitClient (Work in progress)

YuliyaSheyko:


portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
```diff
@@ -33,8 +33,6 @@ trait UserEntry {
 
   def addUserAction(user: User) = (users.returning(users.map(_.userId))) += user
 
-  def deactivateUserByEmail(email: String) = users.filter(_.email === email).map(_.active).update(false)
```
ilyaonishenko: Why it's removed?

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
```diff
@@ -12,9 +12,10 @@ trait ProjectEntry {
     def projectId = column[ProjectId]("project_id", O.PrimaryKey)
     def ownerId = column[UserId]("owner_id")
     def name = column[String]("name")
+    def description = column[String]("description")
```
myazinn: Don't forget to make a migration that will add this column to a DB table.

portal/src/main/scala/cromwell/pipeline/service/GitClient.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import java.io.File
+
+import cromwell.pipeline.datastorage.dto.Project
+
+trait GitClient {
+  def updateFile(project: Project, file: File)
+  def updateListOfFiles(project: Project, files: List[File])
+  def createRepository(project: Project)
+  def getListOfFiles(project: Project, file: File)
```
myazinn: What is this method supposed to do? Why does it require a `File`?

myazinn: Please, specify the return types. Without it, the compiler thinks that the return type is `Unit`.

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
YuliyaSheyko: @ilyaonishenko I thought it would be better to remove a deactivationByEmail feature completely but last time forgot about this method in UserEntry. I'm not sure if we'll need it in the future

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
YuliyaSheyko: @myazinn I'm keeping this in mind. Cause ProjectService test requires an additional changeset with the description column otherwise it fails

portal/src/main/scala/cromwell/pipeline/service/GitClient.scala
YuliyaSheyko: @myazinn 
updateFile is supposed to update a file in user's repository on Gitlab,
getListOfFiles is supposed to return all the files from user's Gitlab project. Probably File won't be java.io.File, maybe it would be better to use a case class. Of that I'm not sure it's just a concept so far.

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
ilyaonishenko: Ok, good.

For the future: best practice is to do all changes in its own branch or commit. Removing user methods should not be done in `gitClient` branch

portal/src/main/scala/cromwell/pipeline/service/GitClient.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import scala.concurrent.Future
+
+trait GitClient {
+
+  def updateFile(project: Project, file: File): Future[Option[File]]
+  def updateListOfFiles(project: Project, files: List[File]): Future[Option[List[File]]]
+  def createRepository(project: Project): Future[Option[Project]]
+  def getListOfFiles(project: Project, file: List[File]): Future[Option[List[File]]]
+}
```
ilyaonishenko: We will need a method to get versions of the project. And get versions of the particular file.

portal/src/main/scala/cromwell/pipeline/datastorage/dao/entry/UserEntry.scala
YuliyaSheyko: Ok, I've got it.
Looks logical

portal/src/main/scala/cromwell/pipeline/service/GitClient.scala
```diff
@@ -1,15 +1,25 @@
...
+  def getListOfFiles(project: Project, path: Path): AsyncResult[List[String]]
+  def getVersionsOfProject(project: Project): AsyncResult[Project]
+  def getVersionsOfFiles(project: Project, path: Path): AsyncResult[List[Version]]
+
+  def getFileTree(project: Project, version: Option[Version] = None): AsyncResult[List[String]]
+
+  def getFile(project: Project, path: Path, version: Option[Version] = None): AsyncResult[String]
 }
+
+case class VersioningException(message: String) extends Exception(message)
```
aborg0: Do you think it is worth extending [NoStackTrace](https://www.scala-lang.org/api/2.12.4/scala/util/control/NoStackTrace.html) too?

YuliyaSheyko: It's just a stub so far but I think it's a good thing to add.

portal/src/main/scala/cromwell/pipeline/service/GitClient.scala
```diff
@@ -0,0 +1,25 @@
...
+trait ProjectVersioning[E >: VersioningException] {
+
+  type AsyncResult[T] = Future[Either[E, T]]
+
+  def updateFile(project: Project, path: Path, content: String): AsyncResult[String]
+  def updateListOfFiles(project: Project, path: Path, content: List[String]): AsyncResult[List[String]]
+  def createRepository(project: Project, path: Path): AsyncResult[Project]
+  def getListOfFiles(project: Project, path: Path): AsyncResult[List[String]]
+  def getVersionsOfProject(project: Project): AsyncResult[Project]
+  def getVersionsOfFiles(project: Project, path: Path): AsyncResult[List[Version]]
```
mnesvit: should be ```getVersionsOfFile``` ?

PR #33, opened 2020-02-11
Epmlstrcmw 119

Benderbej:
This is draft request, thats why contains drafts only. After analyzing, task it will be cleaned(squashed & prepared) and transformed into regular PR.


Benderbej:
What is useful for now - you can download, try and compare both  wdl and womtool jars from here and look how docs can be represented

PR #37, opened 2020-03-02
Epmlstrcmw 133_2

YuliyaSheyko:
Add logger and stop parallel execution in tests


ilyaonishenko:
Considering current status, I propose to add main log config (because we will need it anyway) in the `portal` module. It should take care of stdout and logging in file.

Also something I wanted to say. Such repetition of log conf doesn't look very nice. And however it's the lesser evil, I can see that we can reduce the repetition.

- Main log configuration addition will take care of integration tests, so we can remove copy-pasted log from `it`.  
- We don't start `testcontainers` in `controllers`, `services`, `womtool` and `utils`  so why we're adding conf log there?

YuliyaSheyko:
> Considering current status, I propose to add main log config (because we will need it anyway) in the `portal` module. It should take care of stdout and logging in file.
> 
> Also something I wanted to say. Such repetition of log conf doesn't look very nice. And however it's the lesser evil, I can see that we can reduce the repetition.
> 
> * Main log configuration addition will take care of integration tests, so we can remove copy-pasted log from `it`.
> * We don't start `testcontainers` in `controllers`, `services`, `womtool` and `utils`  so why we're adding conf log there?

- This is a configuration file for tests. Our modules are independent in terms of tests. So I think it’s better to add config in each module because in the future it’ll be possible to make detailed settings of the logger for each module. So far config file is the same, but as the number of tests increases in each module, another setting may be needed.
- I think that we need logger in controllers, services, womtool and utils modules because the root of evil is not in test containers but in liquibase. It gives lots of logs. Unit tests contain not that much logs as itTests but anyway liquibase gives a lot of logs in it. So i think it would be better to fix it.
- I’m not sure about duplicate test configs if it’s a bad decision because anyway they won’t be packed into jar.
- Also I’m not sure about adding test log configiration to the mail log file because if we mix everything in the same file it can be difficult to understand what is used and where, or what is outdated and can be removed.

PR #40, opened 2020-03-06
fix: remove containers initialization outside `beforeAll` method

myazinn:
fix: close connection to the database in case of failed migration
fix: use portNumber from `application.conf`
style: remove redundant braces
style: reformat some code in accordance with scalafmt
style: optimize imports everywhere


Dmitrii-Shevchenko (approved):
Good work, it looks much better!

peter-bonta-epam (approved):
Thank you for making these changes!

OstapStrashevskii (approved):
Good work! If that is demanded by properly tuned scalafmt, I guess most of us need to check it out

iliaonishchenko (approved):

PR #41, opened 2020-03-10
Epmlstrcmw 134

Dmitrii-Shevchenko:
feat: add SBT native container

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -47,12 +47,17 @@ lazy val datasource = project
 
 lazy val portal = project
   .configs(IntegrationTest)
+  .enablePlugins(JavaAppPackaging)
   .settings(
     name := "Portal",
     commonSettings,
     libraryDependencies ++= akkaDependencies ++ jsonDependencies,
     Defaults.itSettings,
     Seq(parallelExecution in Test := false),
+    mappings in Universal ++= Seq(
+      (resourceDirectory in Compile).value / "application.conf" -> "conf/application.conf",
+      (resourceDirectory in Compile).value / "logback.xml" -> "conf/logback.xml"
+    ),
     addCommandAlias("testAll", "; test ; it:test")
   )
   .aggregate(repositories, services, controllers, utils)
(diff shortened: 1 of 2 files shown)


portal/build.sbt
```diff
@@ -0,0 +1,5 @@
+enablePlugins(JavaServerAppPackaging)
```
myazinn: Why do you think it is better to use `JavaServerAppPackaging` instead of `JavaAppPackaging`?

portal/build.sbt
```diff
@@ -0,0 +1,5 @@
...
+mappings in Universal ++= Seq(
+  (baseDirectory in Compile).value / "src/main/resources/application.conf" -> "conf/application.conf",
+  (baseDirectory in Compile).value / "src/main/resources/logback.xml" -> "conf/logback.xml"
+)
```
myazinn: I think we don't have to have two `build.sbt` files in our project. You should configure native packager in the existing `build.sbt`.

portal/build.sbt
```diff
@@ -0,0 +1,5 @@
+enablePlugins(JavaServerAppPackaging)
+mappings in Universal ++= Seq(
+  (baseDirectory in Compile).value / "src/main/resources/application.conf" -> "conf/application.conf",
+  (baseDirectory in Compile).value / "src/main/resources/logback.xml" -> "conf/logback.xml"
```
myazinn: `resourceDirectory` will be more appropriate here than `baseDirectory`.

portal/build.sbt
Dmitrii-Shevchenko: I think that you are right, we don't need the features of server archetype yet. Thank you!

portal/build.sbt
Dmitrii-Shevchenko: Exactly!

portal/build.sbt
Dmitrii-Shevchenko: I don’t know for sure, but I think @ilyaonishenko can describe the reason

ilyaonishenko: I believe it's the style thing. 
If we're going to to continue with single set style, we can keep these changes in the global `build.sbt`

@Dmitrii-Shevchenko you can move it to the main `build.sbt`, but apply this changes **specifically** to the `portal` project.
It will look like 
```
lazy val portal = project
   ...
  .enablePlugins(JavaAppPackaging)
  ...
  .settings(
    ...
    mappings..
  )
```

Dmitrii-Shevchenko: Done

PR #43, opened 2020-03-16
IS-01: Add HttpClient and ProjectVersioning

ilyaonishenko:
It's a first drop of these entities. Most probably they will evolve in future, but now we need them to provide continuous work on the project.


myazinn:
Spectacular work!

PR #49, opened 2020-03-19
EPMLSTRCMW-131. Add deleteProject and updateProject

nikita0levchenko:
Create deleteProjectById and updateProject method's and add test's to them.
https://jirapct.epam.com/jira/browse/EPMLSTRCMW-131

### DIFF
--- a/controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
@@ -2,23 +2,18 @@ package cromwell.pipeline.controller
 
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.testkit.ScalatestRouteTest
-import cromwell.pipeline.datastorage.dao.repository.utils.TestProjectUtils
-import cromwell.pipeline.datastorage.dto.{Project, User, UserId}
+import cromwell.pipeline.datastorage.dao.repository.utils.{ TestProjectUtils, TestUserUtils }
+import cromwell.pipeline.datastorage.dto.{ Project, ProjectDeleteRequest, ProjectUpdateRequest, User, UserId }
 import cromwell.pipeline.datastorage.utils.auth.AccessTokenContent
 import cromwell.pipeline.service.ProjectService
-import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
 import org.mockito.Mockito.when
-import org.scalatest.{AsyncWordSpec, Matchers}
+import org.scalatest.{ AsyncWordSpec, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
 
 import scala.concurrent.Future
 
-class ProjectControllerTest
-  extends AsyncWordSpec
-    with Matchers
-    with ScalatestRouteTest
-    with MockitoSugar
-    with PlayJsonSupport {
+class ProjectControllerTest extends AsyncWordSpec with Matchers with ScalatestRouteTest with MockitoSugar {
 
   private val projectService = mock[ProjectService]
   private val projectController = new ProjectController(projectService)
@@ -37,6 +32,66 @@ class ProjectControllerTest
         Get("/projects?name=" + projectByName) ~> projectController.route(accessToken) ~> check {
           status shouldBe StatusCodes.OK
           responseAs[Option[Project]] shouldEqual (getProjectByNameResponse)
+        }
+      }
+    }
+
+    "delete project by id" should {
+      "return deactivated project entity" in {
+        val dummyProject = TestProjectUtils.getDummyProject()
+        val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId.value)
+        val deactivatedProject = dummyProject.copy(active = false)
+        val request = ProjectDeleteRequest(dummyProject.projectId)
+
+        when(projectService.deactivateProjectById(dummyProject.projectId, UserId(accessToken.userId)))
+          .thenReturn(Future.successful(Some(deactivatedProject)))
+
+        Delete("/projects", request) ~> projectController.route(accessToken) ~> check {
+          responseAs[Project] shouldBe deactivatedProject
+        }
+
+      }
+
+      "return server error if project deactivation was failed" taggedAs (Controller) in {
+        val userId = TestUserUtils.getDummyUserId
+        val request = ProjectDeleteRequest(TestProjectUtils.getDummyProject().projectId)
+        val accessToken = AccessTokenContent(userId.value)
+
+        when(projectService.deactivateProjectById(request.projectId, userId))
+          .thenReturn(Future.failed(new RuntimeException("Something wrong")))
+
+        Delete("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.InternalServerError
+        }
+      }
+    }
+
+    "update project" should {
+      "return status code NoContent if the update was successful" in {
+        val userId = TestUserUtils.getDummyUserId
+        val accessToken = AccessTokenContent(userId.value)
+        val dummyProject = TestProjectUtils.getDummyProject()
+        val request = ProjectUpdateRequest(dummyProject.projectId, dummyProject.name, dummyProject.repository)
+        val updatedProject = dummyProject.copy(name = request.name, repository = request.repository)
+
+        when(projectService.updateProject(request, userId)).thenReturn(Future.successful(1))
+
+        Put("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.NoContent
+        }
+      }
+
+      "return InternalServerError status if the update failed" in {
+        val userId = TestUserUtils.getDummyUserId
+        val accessToken = AccessTokenContent("0")
+        val dummyProject = TestProjectUtils.getDummyProject()
+        val request = ProjectUpdateRequest(dummyProject.projectId, dummyProject.name, dummyProject.repository)
+
+        when(projectService.updateProject(request, userId))
+          .thenReturn(Future.failed(new RuntimeException("Something wrong")))
+
+        Put("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.InternalServerError
         }
       }
     }
--- a/portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
+++ b/portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
@@ -4,18 +4,18 @@ import java.util.UUID
 
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.testkit.ScalatestRouteTest
-import com.dimafeng.testcontainers.{ForAllTestContainer, PostgreSQLContainer}
+import com.dimafeng.testcontainers.{ ForAllTestContainer, PostgreSQLContainer }
 import com.typesafe.config.Config
 import cromwell.pipeline.ApplicationComponents
-import cromwell.pipeline.datastorage.dao.repository.utils.{TestProjectUtils, TestUserUtils}
-import cromwell.pipeline.datastorage.dto.{Project, ProjectId, UserId}
+import cromwell.pipeline.datastorage.dao.repository.utils.{ TestProjectUtils, TestUserUtils }
+import cromwell.pipeline.datastorage.dto.{ Project, ProjectDeleteRequest, ProjectId, ProjectUpdateRequest, UserId }
 import cromwell.pipeline.datastorage.utils.auth.AccessTokenContent
 import cromwell.pipeline.utils.TestContainersUtils
 import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport
-import org.scalatest.{AsyncWordSpec, Matchers}
+import org.scalatest.{ AsyncWordSpec, Matchers }
 
 class ProjectControllerItTest
-  extends AsyncWordSpec
+    extends AsyncWordSpec
     with Matchers
     with ScalatestRouteTest
     with PlayJsonSupport
@@ -26,7 +26,7 @@ class ProjectControllerItTest
   implicit val config: Config = TestContainersUtils.getConfigForPgContainer(container)
   private val components: ApplicationComponents = new ApplicationComponents()
   override protected def beforeAll(): Unit = components.datastorageModule.pipelineDatabaseEngine.updateSchema()
-
+  private val dummyUser = TestUserUtils.getDummyUser()
   import components.controllerModule.projectController
   import components.datastorageModule.projectRepository
   import components.datastorageModule.userRepository
@@ -34,8 +34,7 @@ class ProjectControllerItTest
   "ProjectController" when {
     "getProjectByName" should {
       "return a project with the same name" in {
-        val dummyUser = TestUserUtils.getDummyUser("1")
-        val dummyProject = TestProjectUtils.getDummyProject(ProjectId(UUID.randomUUID().toString), UserId("1"))
+        val dummyProject = TestProjectUtils.getDummyProject(ownerId = dummyUser.userId)
         val projectByNameRequest = dummyProject.name
 
         userRepository
@@ -52,5 +51,38 @@ class ProjectControllerItTest
           )
       }
     }
+
+    "updateProject" should {
+      "return status code NoContend if project was successfully updated" in {
+        val dummyProject = TestProjectUtils.getDummyProject(ownerId = dummyUser.userId)
+        val request = ProjectUpdateRequest(dummyProject.projectId, dummyProject.name, dummyProject.repository)
+        projectRepository
+          .addProject(dummyProject)
+          .flatMap(
+            _ =>
+              projectRepository.updateProject(dummyProject).map { _ =>
+                val accessToken = AccessTokenContent(dummyUser.userId.value)
+                Put("/projects", request) ~> projectController.route(accessToken) ~> check {
+                  status shouldBe StatusCodes.NoContent
+                }
+              }
+          )
+      }
+    }
+
+    "deleteProjectById" should {
+      "return project's entity with false value if project was successfully deactivated" in {
+        val dummyProject = TestProjectUtils.getDummyProject(ownerId = dummyUser.userId)
+        val request = ProjectDeleteRequest(dummyProject.projectId)
+        val deactivatedProjectResponse = dummyProject.copy(active = false)
+        projectRepository.addProject(dummyProject).map { _ =>
+          val accessToken = AccessTokenContent(dummyUser.userId.value)
+          Delete("/projects", request) ~> projectController.route(accessToken) ~> check {
+            responseAs[Project] shouldBe deactivatedProjectResponse
+          }
+        }
+      }
+    }
+
   }
 }
--- a/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
+++ b/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
@@ -5,14 +5,14 @@ import java.util.UUID
 import cromwell.pipeline.datastorage.dto.{ Project, ProjectId, UserId }
 
 object TestProjectUtils {
-  val userUuid: String = UUID.randomUUID().toString
-  val projectUuid: String = UUID.randomUUID().toString
 
+  private def randomUuidStr: String = UUID.randomUUID().toString
+  def getDummyProjectId: ProjectId = ProjectId(randomUuidStr)
   def getDummyProject(
-    projectId: ProjectId = ProjectId(projectUuid),
-    ownerId: UserId = UserId(userUuid),
-    name: String = s"project-$projectUuid",
-    repository: String = s"repo-$projectUuid",
+    projectId: ProjectId = getDummyProjectId,
+    ownerId: UserId = TestUserUtils.getDummyUserId,
+    name: String = s"project-" + randomUuidStr,
+    repository: String = s"repo-" + randomUuidStr,
     active: Boolean = true
   ): Project =
     Project(projectId, ownerId, name, repository, active)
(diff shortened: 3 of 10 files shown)


controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
```diff
@@ -40,5 +39,64 @@ class ProjectControllerTest
...
+          .thenReturn(Future.failed(new RuntimeException("Something wrong")))
+
+        Delete("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.InternalServerError
+        }
+      }
+    }
+
+    "update project" should {
+      "return status code NoContent if " in {
```
myazinn: It looks like an incomplete sentence.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
```diff
@@ -40,5 +39,64 @@ class ProjectControllerTest
...
+        val request = ProjectUpdateRequest(dummyProject.projectId, dummyProject.name, dummyProject.repository)
+
+        when(projectService.updateProject(request, userId)).thenReturn(Future.successful(1))
+
+        Put("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.NoContent
+        }
+      }
+
+      "return InternalServerError status if project id doesn't match" in {
```
myazinn: How does this test checks that the project id doesn't match?

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,36 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
...
 
-  def deactivateProjectById(projectId: ProjectId): Future[Option[Project]] =
-    for {
-      _ <- projectRepository.deactivateProjectById(projectId)
-      getProject <- projectRepository.getProjectById(projectId)
-    } yield getProject
+  def deactivateProjectById(projectId: ProjectId, userId: UserId): Future[Option[Project]] = {
+    val result = projectRepository.getProjectById(projectId)
+    result.flatMap {
+      case Some(project) if project.ownerId != userId => throw new ProjectAccessDeniedException
```
myazinn: Do not throw an exception, return `Future.failed` instead. Same for all other places in this PR.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
```diff
@@ -52,5 +52,36 @@ class ProjectControllerItTest
           )
       }
     }
+
+    "updateProject" should {
+      "return status code NoContend if project was successfully updated" in {
+        val dummyProject = TestProjectUtils.getDummyProject(ProjectId(UUID.randomUUID().toString),UserId(ownerId))
```
myazinn: Why do you need to specify `ProjectId` here? `getDummyProject` can take care of parameters that you don't care about.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,36 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
...
+            project =>
+              if (project.ownerId == userId)
+                projectRepository.updateProject(
+                  project.copy(name = request.name, repository = request.repository)
+                )
+              else Future.failed(new ProjectAccessDeniedException)
+          ) match {
+            case Some(value) => value
+            case None        => Future.failed(new ProjectNotFoundException)
+          }
```
myazinn: It is really hard to read this code. I believe it can be simplified by using pattern matching on the `projectOpt`.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,36 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
...
+      case Some(_) =>
+        for {
+          _ <- projectRepository.deactivateProjectById(projectId)
+          getProject <- projectRepository.getProjectById(projectId)
+        } yield getProject
+      case None => throw new ProjectNotFoundException
+    }
+  }
+
+  def updateProject(request: ProjectUpdateRequest, userId: UserId): Future[Int] =
```
myazinn: Why does it return an `Int`? What does this `Int` mean?

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,36 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
     projectRepository.addProject(project)
   }
 
-  def deactivateProjectById(projectId: ProjectId): Future[Option[Project]] =
-    for {
-      _ <- projectRepository.deactivateProjectById(projectId)
-      getProject <- projectRepository.getProjectById(projectId)
-    } yield getProject
+  def deactivateProjectById(projectId: ProjectId, userId: UserId): Future[Option[Project]] = {
+    val result = projectRepository.getProjectById(projectId)
```
myazinn: 1) You already have a separate method `getProjectById` in this `Service`. Why don't you use it?
2) Why do we need a variable with such a meaningful name? :)

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
nikita0levchenko: fixed

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: This Int - is Id of updated project

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: fixed

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
nikita0levchenko: I need order of parameters in this constructor. I need to explicitly specify userId. Without projectId I'll have type mismatch.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: 1) We have a wrapper for the `ProjectId`. If you want to return the `Id` of the updated project, you must use this wrapper.
2) `ProjectId` is a wrapper for a `String`. Therefore, this `Int` is definitely not an `Id` of the updated project.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
myazinn: The sentence is still incomplete.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
myazinn: Nothing has changed. Therefore, nothing is fixed.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
myazinn: You can use named arguments to specify only `userId`.
`TestProjectUtils.getDummyProject(userId = UserId(ownerId))`

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Nothing has changed. Therefore, nothing is fixed.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Nothing has changed. Therefore, nothing is fixed.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Nothing has changed. Therefore, nothing is fixed.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: Sorry, I was wrong. This is number of updated rows. Should I change this to a more convenient format? If so, which one?

myazinn: This is debatable.
You can either return `Future[Unit]` or `Future[Project]`. `Future[Unit]` returns empty `Success` if the project was updated successfully and `Failure` otherwise. `Future[Project]` is similar but instead of empty `Success`, it returns an updated project. Since during the deactivation of a project you are returning the deactivated project back to the user, I think it would be convenient to return the updated project here.
@ilyaonishenko @mnesvit What do you think?

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
nikita0levchenko: I don't know why, but If I do not specify projectId directly, then my tests fall because such a project already exists

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
nikita0levchenko: Sorry, I didn't push. Fixed now

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
nikita0levchenko: Sorry, I didn't push. Fixed now

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: Sorry, I didn't push. Fixed now

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: Sorry, I didn't push. Fixed now

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: Sorry, I didn't push. Fixed now

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
myazinn: Because you did add a project with the same `projectId` in the previous test.
Unfortunately, `getDummyProject` returns projects with the same `projectId` every time you call it.
You should either fix method `getDummyProject` to make it return project with new `projectId` every time, or fix your tests to ensure that they will never affect each other. Fixing `getDummyProject` project is much easier, but you are free to choose the second option if you want :)

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
ilyaonishenko: I agree with @myazinn

nikita0levchenko: fixed

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Sorry, but I don't see any changes. Variable with a meaningless name still exists and you are still using the method  `projectRepository.getProjectById` directly.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,30 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
...
+    }
+  }
+
+  def updateProject(request: ProjectUpdateRequest, userId: UserId): Future[Option[Project]] =
+    projectRepository
+      .getProjectById(request.projectId)
+      .flatMap {
+        case Some(project) => if (project.ownerId == userId) {
+          projectRepository.updateProject(project.copy(name = request.name, repository = request.repository))
+          projectRepository.getProjectById(project.projectId)
```
myazinn: It is not guaranteed that you will get a project after update since `projectRepository.updateProject` returns a `Future`.  
It is even possible that `projectRepository.updateProject` will fail, but won't notice it. 
@ilyaonishenko you think that it is "good enough" or just didn't notice it? :)

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,11 +34,30 @@ class ProjectService(projectRepository: ProjectRepository)(implicit executionCon
...
+        } else Future.failed(new ProjectAccessDeniedException)
+        case None => Future.failed(new ProjectNotFoundException)
+        case _ => Future.failed(new RuntimeException("Something wrong"))
```
myazinn: Is there any way for us to get here?

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -9,7 +9,7 @@ object TestProjectUtils {
   val projectUuid: String = UUID.randomUUID().toString
 
   def getDummyProject(
-    projectId: ProjectId = ProjectId(projectUuid),
+    projectId: ProjectId = ProjectId(UUID.randomUUID().toString),
```
myazinn: Unfortunately, now there is a duplication of logic in how a `ProjectId` is created for tests. It would be nice if you will find a way to fix it.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
ilyaonishenko: I didn't notice that. 
You're right, this needs to be fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
ilyaonishenko: 😕

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: fixed

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: Discussed today on daily.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
nikita0levchenko: fixed

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
myazinn: You made it even worse :) Now the logic is duplicated everywhere. We should avoid such things.

nikita0levchenko: Then I have no idea how to fix this, because for my test's I need non-constant ProjectId. I can get non-static ProjectId if I will create him in getDummyProject(), not in object. But, unfortunately, this is duplication of logic. I suggest in tests ,that need different ProjectId, create it and pass it to the parameters explicitly, as I had from the very beginning.

myazinn:
```
object TestProjectUtils {

  def getDummyProjectId: ProjectId = ProjectId(randomUuidStr)

  def getDummyProject(
    projectId: ProjectId = getDummyProjectId,
    ownerId: UserId = getDummyUserId,
    name: String = s"project-$randomUuidStr",
    repository: String = s"repo-$randomUuidStr",
    active: Boolean = true
  ): Project =
    Project(projectId, ownerId, name, repository, active)

  // todo move this method to the TestUserUtils
  private def getDummyUserId: UserId = UserId(randomUuidStr)

  private def randomUuidStr: String = UUID.randomUUID.toString
}
}
```
Notice that fields `name` and `repository` contain different `UUID`s than `projectId`. This is okay since `name` and `repository` can be arbitrary and may not contain `projectId` in general. They must be unique and that's the only requirement for them at the moment.
The above code also introduces the method `getDummyProjectId` that will allow anyone to generate `projectId` without knowing how to do this.

nikita0levchenko: Got it.

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -5,14 +5,15 @@ import java.util.UUID
 import cromwell.pipeline.datastorage.dto.{ Project, ProjectId, UserId }
 
 object TestProjectUtils {
-  val userUuid: String = UUID.randomUUID().toString
-  val projectUuid: String = UUID.randomUUID().toString
 
+  private def randomUuidStr: String = UUID.randomUUID().toString
+  def getDummyProjectId: ProjectId = ProjectId(randomUuidStr)
+  private def getDummyUserId: UserId = UserId(randomUuidStr)
```
myazinn: Please, do not remove `// todo` that I left in the previous comment.
This method must be moved to `TestUserUtils` and without that `todo` we can easily forget about it.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
```diff
@@ -40,5 +35,65 @@ class ProjectControllerTest
...
+
+        when(projectService.updateProject(request, userId)).thenReturn(Future.successful(1))
+
+        Put("/projects", request) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.NoContent
+        }
+      }
+
+      "return InternalServerError status if the update failed" in {
+        val userId = TestUserUtils.getDummyUser().userId
```
myazinn: Lines 42, 56, 71 and 85 in this file are good examples of why we should add method `getDummyUserId` into `TestUserUtils`.

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -5,14 +5,15 @@ import java.util.UUID
 import cromwell.pipeline.datastorage.dto.{ Project, ProjectId, UserId }
 
 object TestProjectUtils {
-  val userUuid: String = UUID.randomUUID().toString
-  val projectUuid: String = UUID.randomUUID().toString
 
+  private def randomUuidStr: String = UUID.randomUUID().toString
+  def getDummyProjectId: ProjectId = ProjectId(randomUuidStr)
+  private def getDummyUserId: UserId = UserId(randomUuidStr) // todo move this method to the TestUserUtils
```
myazinn: Since you did add `getDummyUserId` to a `TestUserUtils` class, there is no need in this method.
`TestUserUtils.getDummyUserId` on line 14 and that's all.

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestUserUtils.scala
```diff
@@ -9,8 +9,10 @@ import cromwell.pipeline.utils.StringUtils
 object TestUserUtils {
   val userPassword = "-Pa$$w0rd-"
 
+  private def randomUuidStr: String = UUID.randomUUID().toString
+  def getDummyUserId = UserId(randomUuidStr)
   def getDummyUser(
-    uuid: String = UUID.randomUUID().toString,
+    uuid: String = randomUuidStr,
```
myazinn: It looks like this parameter is unused now. No matter what, you are calling `getDummyUserId` on line 25, which ignores this `uuid` parameter.
If you want to do it right, you must change the signature of this method. Instead of `uuid: String` it must accept `userId: UserId` with the default value `getDummyUserId`. Then this `userId` should be used on line 25.

nikita0levchenko: fixed

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
nikita0levchenko: fixed

PR #50, opened 2020-03-19
EPMLSTRCMW 136: Add implementation of HttpClient

GrigoriiBerezin:


services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
+package cromwell.pipeline.service
```
Dmitrii-Shevchenko: Why did you choose the service module for this?

GrigoriiBerezin: Trait was here and I consulted with Masha about it

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+
+  private def parseHeaders(headers: Map[String, String]): List[RawHeader] =
+    headers.foldLeft(List.empty[RawHeader]) {
+      case (acc, (param, value)) => acc ++ List(RawHeader(param, value))
+    }
+
+  private def queryBuilder(url: String, params: Map[String, String]): String = {
+    val confirmParams = params.map { case (param, value) => s"$param=$value" }
```
OstapStrashevskii: use "collect" instead of "map", to avoid matcherror, when working with partial functions ( @myazinn @ilyaonishenko, please confirm or reject this, I need your advice here )

OstapStrashevskii (changes requested):
Nice work! I leave some comments for you to observe

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
myazinn: This is a good advice in general.
However, in that case it may only confuse others. Can you give an example of the `params` ​​values when this function will throw MatchError?

OstapStrashevskii: may be, this PartialFunction could be more complex, for example, it would not be defined at string.length >= 255, or with Strings which contains special characters, delimiters etc with some regex. We can intercept this mismatch here allready

services/src/test/scala/cromwell/pipeline/service/ControllerClassTest.scala
```diff
@@ -0,0 +1,77 @@
...
+          mockHttp.singleRequest(
+            any[HttpRequest],
+            any[HttpsConnectionContext],
+            any[ConnectionPoolSettings],
+            any[LoggingAdapter]
+          )
+        ).thenReturn(response)
+
+        val client = new TestControllerClient
+        client.put("https://test.com/put", payload = "test payload").flatMap(_ shouldBe cromwellResponse)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/test/scala/cromwell/pipeline/service/ControllerClassTest.scala
```diff
@@ -0,0 +1,77 @@
...
+          mockHttp.singleRequest(
+            any[HttpRequest],
+            any[HttpsConnectionContext],
+            any[ConnectionPoolSettings],
+            any[LoggingAdapter]
+          )
+        ).thenReturn(response)
+
+        val client = new TestControllerClient
+        client.post("https://test.com/put", payload = "test payload").flatMap(_ shouldBe cromwellResponse)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
...
+    url: String,
+    params: Map[String, String] = Map(),
+    headers: Map[String, String] = Map(),
+    payload: String = ""
+  )(implicit ec: ExecutionContext): Future[Response] = {
+    val request = HttpRequest(method = HttpMethods.POST, uri = queryBuilder(url, params))
+      .withEntity(payload)
+      .withHeaders(parseHeaders(headers))
+    val futureResponse: Future[HttpResponse] = http.singleRequest(request)
+
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
...
+  override def put(
+    url: String,
+    params: Map[String, String] = Map(),
+    headers: Map[String, String] = Map(),
+    payload: String = ""
+  )(implicit ec: ExecutionContext): Future[Response] = {
+    val request = HttpRequest(method = HttpMethods.PUT, uri = queryBuilder(url, params))
+      .withEntity(payload)
+      .withHeaders(parseHeaders(headers))
+    val futureResponse: Future[HttpResponse] = http.singleRequest(request)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
...
+  private implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  def http: HttpExt = Http()
+
+  override def get(url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
+    implicit ec: ExecutionContext
+  ): Future[Response] = {
+    val request =
+      HttpRequest(method = HttpMethods.GET, uri = queryBuilder(url, params)).withHeaders(parseHeaders(headers))
+    val futureResponse: Future[HttpResponse] = http.singleRequest(request)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  private def parseHeaders(headers: Map[String, String]): List[RawHeader] =
+    headers.foldLeft(List.empty[RawHeader]) {
+      case (acc, (param, value)) => acc ++ List(RawHeader(param, value))
+    }
+
+  private def queryBuilder(url: String, params: Map[String, String]): String = {
+    val confirmParams = params.map { case (param, value) => s"$param=$value" }
+    confirmParams match {
+      case p :: ps => url + "?" + confirmParams.mkString("&")
```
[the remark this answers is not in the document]
GrigoriiBerezin: I think that can be hardcoded because of private method that is used only inside class and there is no need to change delimiters

services/src/test/scala/cromwell/pipeline/service/ControllerClassTest.scala
```diff
@@ -0,0 +1,77 @@
...
+          mockHttp.singleRequest(
+            any[HttpRequest],
+            any[HttpsConnectionContext],
+            any[ConnectionPoolSettings],
+            any[LoggingAdapter]
+          )
+        ).thenReturn(response)
+
+        val client = new TestControllerClient
+        client.get("https://test.com/get").flatMap(_ shouldBe cromwellResponse)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Done, thx

services/src/main/scala/cromwell/pipeline/service/ControllerClient.scala
GrigoriiBerezin: Params are Map of Strings, so no matcherror is likely, but thank you for advice, I will take it.

Dmitrii-Shevchenko (approved):
Great work!

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -1,30 +1,31 @@
...
+import cromwell.pipeline.service.Response
 import org.mockito.Matchers.any
 import org.mockito.Mockito.when
 import org.scalatest.{ AsyncWordSpec, BeforeAndAfterAll, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
 
 import scala.concurrent.{ ExecutionContext, Future }
 
-class ControllerClassTest extends AsyncWordSpec with Matchers with MockitoSugar with BeforeAndAfterAll {
+class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with BeforeAndAfterAll {
```
IaroslavTavchenkov: BeforeAndAfterAll seems like unused feature

IaroslavTavchenkov (approved):

OstapStrashevskii (approved):
Good work!

nikita0levchenko (approved):
Good work!

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+      HttpRequest(method = HttpMethods.GET, uri = queryBuilder(url, params)).withHeaders(parseHeaders(headers))
+    )
+    responsify(futureResponse)
+  }
+
+  override def post(
+    url: String,
+    params: Map[String, String] = Map(),
+    headers: Map[String, String] = Map(),
+    payload: String = ""
```
ilyaonishenko: Empty string is never a good thing. 
@myazinn come here and tell them the truth, please. 

Also, it's a good idea to make slides for newcomers like "5 simple things you should never do".

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+    val futureResponse: Future[HttpResponse] = http.singleRequest(
+      HttpRequest(method = HttpMethods.PUT, uri = queryBuilder(url, params))
+        .withEntity(payload)
+        .withHeaders(parseHeaders(headers))
+    )
+    responsify(futureResponse)
+  }
+
+  private def responsify(futureResponse: Future[HttpResponse])(implicit ec: ExecutionContext): Future[Response] =
+    futureResponse.flatMap { response: HttpResponse =>
```
ilyaonishenko: `: HttpResponse` part is redundant.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+      response.entity.dataBytes.runFold(ByteString(""))(_ ++ _).map { body =>
```
ilyaonishenko: I'm pretty sure this is unnecessary, in this case we can use just `response.entity.strict(timeout)`.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+        val responseBody = body.decodeString(Charset.defaultCharset())
+        Response(
+          response.status.intValue(),
+          responseBody,
+          response.headers.foldLeft(Map[String, List[String]]()) {
+            case (acc, header) =>
+              acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
+          }
```
ilyaonishenko: If I understood correctly, you're merging headers with the same name in the one header here.
I'm not sure it's a good idea, we're working like a proxy in this class, we shouldn't manipulate the headers.

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
ilyaonishenko: Looks like it's true

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with BeforeAndAfterAll {
+  implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
+
+  val mockHttp: HttpExt = mock[HttpExt]
+  val cromwellResponse: Response = Response(200, "", Map())
+
+  class TestAkkaHttpClient extends AkkaHttpClient {
+    override def http: HttpExt = mockHttp
+  }
```
ilyaonishenko: We could use WireMock instead of mocking it in this way

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.scalatest.{ AsyncWordSpec, BeforeAndAfterAll, Matchers }
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with BeforeAndAfterAll {
+  implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
+
+  val mockHttp: HttpExt = mock[HttpExt]
+  val cromwellResponse: Response = Response(200, "", Map())
```
ilyaonishenko: Not a good naming and not a good idea.

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    "get" should {
+      "return OK response status" taggedAs Controller in {
+        val response = Future[HttpResponse](HttpResponse())
+        when(
+          mockHttp.singleRequest(
+            any[HttpRequest],
+            any[HttpsConnectionContext],
+            any[ConnectionPoolSettings],
+            any[LoggingAdapter]
```
ilyaonishenko: We need to be less abstract in tests...

iliaonishchenko (changes requested):
Thank you for your work

There're several things should be fixed..

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.concurrent.{ ExecutionContext, Future }
+
+class AkkaHttpClient extends HttpClient {
+  private implicit val actorSystem: ActorSystem = ActorSystem()
+  private implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  def http: HttpExt = Http()
+
+  override def get(url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
+    implicit ec: ExecutionContext
```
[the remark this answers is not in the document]
GrigoriiBerezin: I know that every request have standard headers, but Akka does it yourself

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+import akka.util.ByteString
+import cromwell.pipeline.service.{ HttpClient, Response }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class AkkaHttpClient extends HttpClient {
+  private implicit val actorSystem: ActorSystem = ActorSystem()
+  private implicit val materializer: ActorMaterializer = ActorMaterializer()
+
+  def http: HttpExt = Http()
```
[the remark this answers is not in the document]
GrigoriiBerezin: I used this definition for Mock tests.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
GrigoriiBerezin: Ok, I get it, but how can I use the same header with different values if it's map?

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  override def get(url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
+    implicit ec: ExecutionContext
+  ): Future[Response] = {
+    val futureResponse: Future[HttpResponse] = http.singleRequest(
+      HttpRequest(method = HttpMethods.GET, uri = queryBuilder(url, params)).withHeaders(parseHeaders(headers))
```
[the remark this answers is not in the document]
GrigoriiBerezin: Removed this method and used withQuery, thanks

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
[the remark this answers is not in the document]
ilyaonishenko: ?

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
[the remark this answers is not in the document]
ilyaonishenko: Please consider adding `WireMock`

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,85 @@
...
+import akka.http.scaladsl.{ Http, HttpExt }
+import akka.stream.ActorMaterializer
+import akka.util.ByteString
+import cromwell.pipeline.service.{ HttpClient, Response }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class AkkaHttpClient extends HttpClient {
+  private implicit val actorSystem: ActorSystem = ActorSystem()
+  private implicit val materializer: ActorMaterializer = ActorMaterializer()
```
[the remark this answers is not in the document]
GrigoriiBerezin: Fixed it, thanks

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
GrigoriiBerezin: Didn't find this dependency in project. Should I add it and stub server for verify client response?

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
GrigoriiBerezin: Change it, thanks

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
GrigoriiBerezin: Will fix it after getting how to work with WireMock

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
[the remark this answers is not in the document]
GrigoriiBerezin: Yep

myazinn: Yep, you should never use such placeholders.
If you think that there are cases when you don't need this parameter (or a client may not have it), then you should take `Option[String]` as a parameter.
Otherwise, this parameter may be passed from one method to another many time causing a lot of corrupted data. Fixing this is usually a nightmare :)

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
GrigoriiBerezin: Fixed, thanks

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
GrigoriiBerezin: Change it, thanks

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
GrigoriiBerezin: Get it, thanks, remove empty string. But I can not change trait, where payload is a String, not Optional[String]

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
GrigoriiBerezin: Removed

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -0,0 +1,158 @@
...
+  implicit val materializer: ActorMaterializer = ActorMaterializer()
+  implicit val ec: ExecutionContext = actorSystem.dispatcher
+
+  val wireMockServer = new WireMockServer(wireMockConfig().dynamicPort())
+
+  override def beforeEach: Unit =
+    wireMockServer.start()
+
+  override def afterEach: Unit =
+    wireMockServer.stop()
```
myazinn: When using methods like `beforeAll`, `beforeEach`, `afterAll`, `afterEach` a good practice is to call corresponding methods from parent. I.e.
```
  override def beforeEach: Unit = {
    super.beforeEach
    wireMockServer.start()
}
```
and
```
override def afterEach: Unit = {
    wireMockServer.stop()
    super.afterEach
}
```
You never know what these methods do. Omitting them may cause some bugs.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,71 @@
...
+      HttpRequest(method = HttpMethods.POST, uri = Uri(url).withQuery(Query(params)))
+        .withEntity(payload)
+        .withHeaders(parseHeaders(headers))
+    )
+    responsify(futureResponse)
+  }
+
+  private def responsify(futureResponse: Future[HttpResponse])(implicit ec: ExecutionContext): Future[Response] =
+    futureResponse.flatMap { response =>
+      response.entity.toStrict(300.millis).map { body =>
```
myazinn: Avoid magic numbers.
You can make it a private class field or you can even take it as an implicit parameter for a whole class, but I'm not sure whether it would be overkill or not.
@ilyaonishenko how do you think?

myazinn (changes requested):

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
ilyaonishenko: I would move it to the configuration file.

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
GrigoriiBerezin: Thank you for your advice, changed it

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -0,0 +1,71 @@
...
+              acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
+          }
+        )
+      }
+    }
+
+  private def parseHeaders(headers: Map[String, String]): List[RawHeader] =
+    headers.foldLeft(List.empty[RawHeader]) {
+      case (acc, (param, value)) => acc ++ List(RawHeader(param, value))
+    }
```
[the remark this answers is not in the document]
GrigoriiBerezin: Thank you for your advice, changed it to map

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
myazinn: Yep, I thought so too. I'm just not sure that it should be done in this PR. Therefore, I proposed options that will make easier the movement of this variable to the config file.

GrigoriiBerezin: I messed up with config

iliaonishchenko (approved):
Looks good to me.

myazinn (approved):
Looks good to me
Now you must squash all commits

PR #61, opened 2020-04-14
EPMLSTRCMW-148: Add endpoint for upload file

GrigoriiBerezin:
Service and Controller for upload file were added.
Did not implement controller in ControllerModule, it was in EPMLSTRCMW-147

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -3,7 +3,7 @@ package cromwell.pipeline.controller
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
-import cromwell.pipeline.datastorage.dto.FileContent
+import cromwell.pipeline.datastorage.dto.{ FileContent, ProjectUpdateFileRequest }
 import cromwell.pipeline.datastorage.utils.auth.AccessTokenContent
 import cromwell.pipeline.service.ProjectFileService
 import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
@@ -19,6 +19,17 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
           entity(as[FileContent]) { request =>
             onComplete(wdlService.validateFile(request)) {
               case Success(Left(e)) => complete(StatusCodes.Conflict, e.errors)
+              case Success(_)       => complete(StatusCodes.OK)
+              case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
+            }
+          }
+        }
+      },
+      path("files") {
+        post {
+          entity(as[ProjectUpdateFileRequest]) { request =>
+            onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
+              case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
               case Success(_)       => complete(StatusCodes.OK)
               case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
             }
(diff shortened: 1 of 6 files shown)


controllers/src/test/scala/cromwell/pipeline/controller/ProjectFileControllerTest.scala
```diff
@@ -0,0 +1,51 @@
...
+        Post("/files", request) ~> projectFileController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.OK
+        }
+      }
+
+      "return InternalServerError for bad request" taggedAs Controller in {
+        val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId.value)
+        val project = TestProjectUtils.getDummyProject()
+        val projectFile = ProjectFile(Paths.get("folder/test.txt"), "file context")
+        val request = ProjectUpdateFileRequest(project, projectFile)
```
Dmitrii-Shevchenko, line 40: I think you don't need create the same val's in every test

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTest.scala
```diff
@@ -0,0 +1,39 @@
...
+        val project = TestProjectUtils.getDummyProject()
+        val projectFile = ProjectFile(Paths.get("test.txt"), "File content")
+
+        when(projectVersioning.updateFile(project, projectFile)).thenReturn(Future.successful(Right("Success")))
+        projectFileService.uploadFile(project, projectFile).map(_ shouldBe Right("Success"))
+      }
+
+      "return error message for error request" taggedAs Service in {
+        val project = TestProjectUtils.getDummyProject()
+        val projectFile = ProjectFile(Paths.get("test.txt"), "File content")
```
Dmitrii-Shevchenko, line 31: Common values for both as above

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTest.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scala.concurrent.Future
+
+class ProjectFileServiceTest extends AsyncWordSpec with Matchers with MockitoSugar {
+
+  private val projectVersioning = mock[ProjectVersioning[VersioningException]]
+  private val projectFileService = new ProjectFileService(projectVersioning)
+
+  "ProjectFileServiceTest" when {
+    // not sure about this tests
```
OstapStrashevskii, line 19: You test not only update file here but projectFileService(35 line), so, that's all right, since you have the only one method in ProjectFileService.

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTest.scala
GrigoriiBerezin: Changed it, thanks

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -24,6 +24,17 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
             }
           }
         }
+      },
+      path("files") {
+        post {
+          entity(as[ProjectUpdateFileRequest]) { request =>
+            onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
+              case Success(Left(e)) => complete(StatusCodes.InternalServerError, e.getMessage)
```
mnesvit, line 32: What exactly does it mean to get ```Left``` here? Is it reasonable to return server error? In case of invalid file code 4xx could be better

GrigoriiBerezin: Yep, my bad. But I don't know what invalid file code I should return exactly because I get only VersioningException with error message

GrigoriiBerezin: Fixed on `IamTeapot`

ilyaonishenko: Please add TODO. This status code should be fixed later, when `updareFile` will return good issues.

GrigoriiBerezin: Done

PR #63, opened 2020-04-20
 EPMLSTRCMW-142: Get project versions form gitlab

AndreiMaksimenko:
Add method to get project versions from GilLab
refactor Project Version class and add Commit class
add Json serialization

### DIFF
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
@@ -52,7 +52,16 @@ object ProjectUpdateRequest {
   implicit val updateRequestFormat: OFormat[ProjectUpdateRequest] = Json.format[ProjectUpdateRequest]
 }
 
-final case class Version(value: String) extends AnyVal
+final case class Version(name: String, message: String, target: String, commit: Commit)
+object Version {
+  implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
+  implicit val versionsPlayListFormat: Reads[Seq[Version]] = Reads.seq(versionPlayFormat)
+}
+
+final case class Commit(id: String)
+object Commit {
+  implicit val commitFormat: OFormat[Commit] = Json.format[Commit]
+}
 
 final case class ProjectFile(path: Path, content: String)
 
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
@@ -10,10 +10,13 @@ final case class Project(
   projectId: ProjectId,
   ownerId: UserId,
   name: String,
-  repository: String,
   active: Boolean,
+  repository: Option[Repository] = None,
   visibility: Visibility = Private
-)
+) {
+  def withRepository(repositoryPath: Option[String]): Project =
+    this.copy(repository = repositoryPath.map(Repository(_)))
+}
 object Project {
   implicit lazy val projectFormat: OFormat[Project] = Json.format[Project]
   implicit lazy val projectWrites: Writes[Project] = (project: Project) =>
@@ -21,8 +24,8 @@ object Project {
       "projectId" -> project.projectId.value,
       "ownerId" -> project.ownerId,
       "name" -> project.name,
-      "repository" -> project.repository,
       "active" -> project.active,
+      "repository" -> project.repository,
       "visibility" -> Visibility.toString(project.visibility),
       "path" -> project.projectId.value
     )
@@ -32,6 +35,12 @@ final case class ProjectId(value: String) extends MappedTo[String]
 
 object ProjectId {
   implicit lazy val projectIdFormat: Format[ProjectId] = implicitly[Format[String]].inmap(ProjectId.apply, _.value)
+}
+
+final case class Repository(value: String) extends MappedTo[String]
+
+object Repository {
+  implicit lazy val repositoryFormat: Format[Repository] = implicitly[Format[String]].inmap(Repository.apply, _.value)
 }
 
 final case class ProjectAdditionRequest(name: String)
@@ -46,7 +55,7 @@ object ProjectDeleteRequest {
   implicit lazy val projectDeleteFormat: OFormat[ProjectDeleteRequest] = Json.format[ProjectDeleteRequest]
 }
 
-final case class ProjectUpdateRequest(projectId: ProjectId, name: String, repository: String)
+final case class ProjectUpdateRequest(projectId: ProjectId, name: String, repository: Option[Repository])
 
 object ProjectUpdateRequest {
   implicit val updateRequestFormat: OFormat[ProjectUpdateRequest] = Json.format[ProjectUpdateRequest]
@@ -55,7 +64,6 @@ object ProjectUpdateRequest {
 final case class Version(name: String, message: String, target: String, commit: Commit)
 object Version {
   implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
-  implicit val versionsPlayListFormat: Reads[Seq[Version]] = Reads.seq(versionPlayFormat)
 }
 
 final case class Commit(id: String)
--- a/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
@@ -37,7 +37,26 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     }
   }
   override def getFiles(project: Project, path: Path)(implicit ec: ExecutionContext): AsyncResult[List[String]] = ???
-  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = ???
+
+  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Seq[Version]] =
+    if (project.repository == null)
+      Future.failed(VersioningException("There is no repository in this project"))
+    else {
+      val versionsListUrl: String = s"${config.url}projects/${project.repository}/repository/tags"
+      httpClient
+        .get(url = versionsListUrl, headers = config.token)
+        .map(
+          resp =>
+            if (resp.status != StatusCodes.OK.intValue)
+              Left(VersioningException(s"Could not take versions. Response status: ${resp.status}"))
+            else {
+              val versionsBody = Json.parse(resp.body).validate[Seq[Version]]
+              Right(versionsBody.get)
+            }
+        )
+        .recover { case e: Throwable => Left(VersioningException(e.getMessage)) }
+    }
+
   override def getFileVersions(project: Project, path: Path)(
     implicit ec: ExecutionContext
   ): AsyncResult[List[Version]] = ???
--- a/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
@@ -1,11 +1,9 @@
 package cromwell.pipeline.service
 import java.nio.file.Path
 
-import cromwell.pipeline.utils.HttpStatusCodes
-
 import cromwell.pipeline.datastorage.dto.{ Project, ProjectFile, Version }
-import cromwell.pipeline.utils.GitLabConfig
-import play.api.libs.json.Json
+import cromwell.pipeline.utils.{ GitLabConfig, HttpStatusCodes }
+import play.api.libs.json.{ JsError, JsResult, JsSuccess, Json }
 
 import scala.concurrent.{ ExecutionContext, Future }
 
@@ -19,8 +17,7 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     implicit ec: ExecutionContext
   ): AsyncResult[List[String]] = ???
 
-  override def createRepository(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = {
-    def projectWithRepository(repositoryFullPath: String): Project = project.copy(repository = repositoryFullPath)
+  override def createRepository(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] =
     if (!project.active)
       Future.failed(VersioningException("Could not create a repository for deleted project."))
     else {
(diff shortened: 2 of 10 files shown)


services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -30,13 +30,34 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
+      httpClient
+        .get(url = versionsListUrl, headers = config.token)
+        .map(
+          resp =>
+            if (resp.status != StatusCodes.OK.intValue)
+              Left(VersioningException(s"Could not take versions. Response status: ${resp.status}"))
+            else {
+              println(resp.body)
+              val versionsBody = Json.parse(resp.body).validate[Seq[Version]]
+              println(versionsBody)
```
Dmitrii-Shevchenko, line 54: Remove println pls

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -17,6 +17,7 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
   val gitLabProjectVersioning: GitLabProjectVersioning = new GitLabProjectVersioning(mockHttpClient, gitLabConfig)
   import ProjectContext._
   "GitLabProjectVersioning" when {
+
```
Dmitrii-Shevchenko, line 20: Remove free line

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -38,19 +39,50 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
...
   object ProjectContext {
     lazy val activeProject: Project = TestProjectUtils.getDummyProject()
     lazy val inactiveProject: Project = activeProject.copy(active = false)
     lazy val noRepoProject: Project = activeProject.copy(repository = null)
-    lazy val projectWithRepo: Project = activeProject.copy(repository = gitLabConfig.idPath + activeProject.projectId)
+    lazy val projectWithRepo: Project =
+      activeProject.copy(repository = gitLabConfig.idPath + activeProject.projectId.value)
+    lazy val dummyVersion: Version = TestProjectUtils.getDummyVersion()
   }
+
```
Dmitrii-Shevchenko, line 87: Same

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -38,19 +39,50 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
...
+      "throw new VersioningException with 400 response" taggedAs Service in {
         when(request(activeProject)).thenReturn(Future.successful(Response(StatusCodes.BadRequest.intValue, "", Map())))
         gitLabProjectVersioning.createRepository(activeProject).map {
           _ shouldBe Left(VersioningException("The repository was not created. Response status: 400"))
         }
       }
     }
+
+    "getProjectVersions" should {
+      def dummyVersionsJson: String = s"[${Json.stringify(Json.toJson(dummyVersion))}]"
```
GrigoriiBerezin, line 51: Why def?

GrigoriiBerezin (changes requested):
Nice done. But I have some questions

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
IaroslavTavchenkov: > Why def?

why don't you like defs ?)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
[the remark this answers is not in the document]
AndreiMaksimenko: If you mean a pattern matching, so I think it will be less readable.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
AndreiMaksimenko: Cause we can crush on the first test, and we will not reach this value, so I think it's better to call a method and do not have a val in memory.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
AndreiMaksimenko: Oh! Fixed.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
GrigoriiBerezin: Than lazy val is more fitting for this

AndreiMaksimenko: Deal! Fixed.

Dmitrii-Shevchenko (approved):
Great work!

OstapStrashevskii (approved):
Nice work!

GrigoriiBerezin (approved):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -52,7 +52,16 @@ object ProjectUpdateRequest {
   implicit val updateRequestFormat: OFormat[ProjectUpdateRequest] = Json.format[ProjectUpdateRequest]
 }
 
-final case class Version(value: String) extends AnyVal
+final case class Version(name: String, message: String, target: String, commit: Commit)
+object Version {
+  implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
+  implicit val versionsPlayListFormat: Reads[Seq[Version]] = Reads.seq(versionPlayFormat)
```
myazinn, line 58: No need for a reader of an object sequence.
If you do have a formatter for a single object, `play-json` must be able to handle other things automatically.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -36,7 +36,26 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     }
   }
   override def getFiles(project: Project, path: Path)(implicit ec: ExecutionContext): AsyncResult[List[String]] = ???
-  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = ???
+
+  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Seq[Version]] =
+    if (project.repository == null)
+      Future.failed(VersioningException("There is no repository in this project"))
```
myazinn, line 42: As we discussed, you should not check it here. If you do have a `Project`, you may assume that all of his fields are filled correctly. Otherwise, it must be fixed by whoever created an invalid object.

myazinn (changes requested):
Good work!

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -36,7 +36,26 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
+      val versionsListUrl: String = s"${config.url}projects/${project.repository}/repository/tags"
+      httpClient
+        .get(url = versionsListUrl, headers = config.token)
+        .map(
+          resp =>
+            if (resp.status != StatusCodes.OK.intValue)
+              Left(VersioningException(s"Could not take versions. Response status: ${resp.status}"))
+            else {
+              val versionsBody = Json.parse(resp.body).validate[Seq[Version]]
+              Right(versionsBody.get)
```
myazinn, line 53: Almost forgot.
You must check whether `versionsBody` contains `JsSuccess` or `JsError` using pattern matching instead of simple `.get`. If it contains `JsError` then you must return `Left` instead of `Right`.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -38,13 +38,41 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
...
+      "throw new VersioningException with 400 response" taggedAs Service in {
         when(request(activeProject)).thenReturn(Future.successful(Response(StatusCodes.BadRequest.intValue, "", Map())))
         gitLabProjectVersioning.createRepository(activeProject).map {
           _ shouldBe Left(VersioningException("The repository was not created. Response status: 400"))
         }
       }
     }
+
+    "getProjectVersions" should {
+      lazy val dummyVersionsJson: String = s"[${Json.stringify(Json.toJson(dummyVersion))}]"
```
myazinn, line 50: You may leave it as is, but I think the following code is a bit better
`Json.stringify(Json.toJson(Seq(dummyVersion))`
Also, no need for a `lazy` modifier.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
AndreiMaksimenko: Fixed. Added creation of project on adding to DB.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
AndreiMaksimenko: Very strange but everything works now. Fixed.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
AndreiMaksimenko: Fixed. But I ask you to check, I use "try" and "as" structures. Is it ok?

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
AndreiMaksimenko: Lazy removed.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
myazinn: Good enough, but in my opinion, you can use `.validate` + `.fold` or `.validate` + pattern matching to handle this situation without throwing an exception. Take a look at the implementation of the `.as` method. You can do a similar thing.

AndreiMaksimenko: Fixed to validate.

myazinn (approved):
Nice work!

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -20,7 +18,8 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
   ): AsyncResult[List[String]] = ???
 
   override def createRepository(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = {
-    def projectWithRepository(repositoryFullPath: String): Project = project.copy(repository = repositoryFullPath)
+    def projectWithRepository(repositoryFullPath: String): Project =
+      project.copy(repository = Some(Repository(repositoryFullPath)))
```
ilyaonishenko, line 22: It should be part of `Project` entity in my opinion. 
So it should be defined inside `Project` case class with the name `withRepository`

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -33,25 +32,48 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
       }
       "return new active Project with 201 response" taggedAs Service in {
         val project = projectWithRepo
-        when(request(activeProject)).thenReturn(Future.successful(Response(StatusCodes.Created.intValue, "", Map())))
+        when(request(activeProject)).thenReturn(Future.successful(Response(HttpStatusCodes.Created, "", Map())))
```
ilyaonishenko, line 35: It's better to create a value for it, for example `emptyBody`  and define it as "". Even better is to add some sense to this definition and give up idea of empty strings usage.

iliaonishchenko (changes requested):
Looks almost ready to me.  
There're some cosmetic changes need to be done.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -36,7 +36,26 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     }
   }
   override def getFiles(project: Project, path: Path)(implicit ec: ExecutionContext): AsyncResult[List[String]] = ???
-  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = ???
+
+  override def getProjectVersions(project: Project)(implicit ec: ExecutionContext): AsyncResult[Seq[Version]] =
+    if (project.repository == null)
```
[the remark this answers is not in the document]
AndreiMaksimenko: It was already fixed!

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
AndreiMaksimenko: Done. Please check, did I understand you right.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
AndreiMaksimenko: Fixed! I made EmptyBody - cause it is constant.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
ilyaonishenko: @AndreiMaksimenko It should take Option[String]. 
In your case if we pass there, for example, `null` it won't be processed nicely. (Because of the `Some` )

myazinn: @ilyaonishenko are you sure that it makes sense here? If `repositoryFullPath` is `null`, then we will have the following
`Option(Repository(null))`
not
`Option(null)`
Therefore, `Option` will not save us anyway.

AndreiMaksimenko: Hmmm... it is strange I think. In that case of null who guarantees that there will not be Some[Null] in that Option - and we will have the same situation.

AndreiMaksimenko: I can add one more method withNoRepositry() - to create empty repo Project.

AndreiMaksimenko: Fixed

iliaonishchenko (approved):

PR #70, opened 2020-04-30
feat: Add validation to the uploadFile endpoint

manvicha:
When a file is upload this is validated. If the
validation failed and the file was upload
successfully then it will return a Status 201.
If the validation was right and the file was
upload successfully then it will return a Status
200

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -27,12 +27,18 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
       },
       path("files") {
         post {
-          entity(as[ProjectUpdateFileRequest]) { request =>
-            onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
-              case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
-              case Success(_)       => complete(StatusCodes.OK)
-              case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
-            }
+          entity(as[ProjectUpdateFileRequest]) {
+            request =>
+              onComplete(wdlService.validateFile(FileContent(request.projectFile.content))) {
+                case Success(value) =>
+                  val validationStatus = value.fold(ve => false, un => true)
+                  onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
+                    case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
+                    case Success(_)       => if (validationStatus) complete(StatusCodes.OK) else complete(StatusCodes.Created)
+                    case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
+                  }
+                case Failure(e) => complete(StatusCodes.InternalServerError, e.getMessage)
+              }
           }
         }
       }
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -40,11 +40,8 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
                   case (Left(_), Left(response))   => StatusCodes.ImATeapot.intValue -> response.message
                 }
               }) {
-                case Success(value) =>
-                  value match {
-                    case tuple @ (status: Int, message: String) => complete(status, message)
-                  }
-                case Failure(e) => complete(StatusCodes.InternalServerError, e.getMessage)
+                case Success((status, message)) => complete(status, message)
+                case Failure(e)                 => complete(StatusCodes.InternalServerError, e.getMessage)
               }
           }
         }
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -29,7 +29,7 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
         post {
           entity(as[ProjectUpdateFileRequest]) {
             request =>
-              val result = for {
+              onComplete(for {
                 validateResponse <- wdlService.validateFile(FileContent(request.projectFile.content))
                 uploadResponse <- wdlService.uploadFile(request.project, request.projectFile, request.version)
               } yield {
@@ -39,11 +39,7 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
                   case (Right(_), Left(response))  => StatusCodes.ImATeapot.intValue -> response.message
                   case (Left(_), Left(response))   => StatusCodes.ImATeapot.intValue -> response.message
                 }
-              }
-
-              while (!result.isCompleted) {}
-
-              result.value.get match {
+              }) {
                 case Success(value) =>
                   value match {
                     case tuple @ (status: Int, message: String) => complete(status, message)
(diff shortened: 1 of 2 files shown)


controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,18 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
+              onComplete(wdlService.validateFile(FileContent(request.projectFile.content))) {
+                case Success(value) =>
+                  val validationStatus = value.fold(ve => false, un => true)
+                  onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
+                    case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
+                    case Success(_)       => if (validationStatus) complete(StatusCodes.OK) else complete(StatusCodes.Created)
+                    case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
+                  }
+                case Failure(e) => complete(StatusCodes.InternalServerError, e.getMessage)
+              }
```
GrigoriiBerezin, line 41: Is it correct that we upload file even if it's not validate?

manvicha: Yes, according with the jira task even if the file is not valid this needs to be uploaded. The response will be Created (201) instead of Success(200)

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,18 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
+            request =>
+              onComplete(wdlService.validateFile(FileContent(request.projectFile.content))) {
+                case Success(value) =>
+                  val validationStatus = value.fold(ve => false, un => true)
+                  onComplete(wdlService.uploadFile(request.project, request.projectFile)) {
+                    case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
+                    case Success(_)       => if (validationStatus) complete(StatusCodes.OK) else complete(StatusCodes.Created)
+                    case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
+                  }
+                case Failure(e) => complete(StatusCodes.InternalServerError, e.getMessage)
```
ilyaonishenko, line 40: In Scala we're using for-comprehensions to sequential or parallel work.
Instead of using `onComplete` two times you should consider using for-comprehensions.

You can read nice explanation here: https://stackoverflow.com/questions/19045936/scalas-for-comprehension-with-futures

manvicha: I have applied the change to use for-comprehensions. However the number of lines have increased and I think that the logic too

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,30 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
+                if (validateResponse.isRight && uploadResponse.isRight) {
+                  StatusCodes.OK.intValue -> uploadResponse.right.get
+                } else if (validateResponse.isLeft && uploadResponse.isRight) {
+                  StatusCodes.Created.intValue -> uploadResponse.right.get
+                } else if (uploadResponse.isLeft) {
+                  StatusCodes.ImATeapot.intValue -> uploadResponse.left.get.message
+                }
+              }
+
+              while (!result.isCompleted) {}
```
mnesvit, line 45: It's better to remove this string and replace ```result.value.get``` with ```onComplete(result)```

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,30 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
+                validateResponse <- wdlService.validateFile(FileContent(request.projectFile.content))
+                uploadResponse <- wdlService.uploadFile(request.project, request.projectFile, request.version)
+              } yield {
+                if (validateResponse.isRight && uploadResponse.isRight) {
+                  StatusCodes.OK.intValue -> uploadResponse.right.get
+                } else if (validateResponse.isLeft && uploadResponse.isRight) {
+                  StatusCodes.Created.intValue -> uploadResponse.right.get
+                } else if (uploadResponse.isLeft) {
+                  StatusCodes.ImATeapot.intValue -> uploadResponse.left.get.message
+                }
```
myazinn, line 42: How about pattern matching?:) It will save us from accidental errors when we check that it's `Right` and the call `.left.get`.
```
  (validateResponse, uploadResponse) match {
    case (Right(_), Right(response)) => ???
    case (Left(_), Right(response)) => ???
    case (Left(_), _) => ???
  }
```

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,30 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
-          entity(as[ProjectUpdateFileRequest]) { request =>
-            onComplete(wdlService.uploadFile(request.project, request.projectFile, request.version)) {
-              case Success(Left(e)) => complete(StatusCodes.ImATeapot, e.getMessage) // TODO: change status code
-              case Success(_)       => complete(StatusCodes.OK)
-              case Failure(e)       => complete(StatusCodes.InternalServerError, e.getMessage)
-            }
+          entity(as[ProjectUpdateFileRequest]) {
+            request =>
+              val result = for {
+                validateResponse <- wdlService.validateFile(FileContent(request.projectFile.content))
```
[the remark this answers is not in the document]
manvicha: The jira task says that even if the validation is not correct the file needs to be uploaded, just to be sure, do i change the logic to not upload the file?

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
manvicha: onComplete function was implemented and line 45 was deleted

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
manvicha: The pattern matching was implemented and _if_ logic was removed

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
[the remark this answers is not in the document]
mnesvit: @manvicha i've checked it once again, sorry, it was my mistake

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -27,12 +27,25 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
...
+                  case (Right(_), Right(response)) => StatusCodes.OK.intValue -> response
+                  case (Left(_), Right(response))  => StatusCodes.Created.intValue -> response
+                  case (Right(_), Left(response))  => StatusCodes.ImATeapot.intValue -> response.message
+                  case (Left(_), Left(response))   => StatusCodes.ImATeapot.intValue -> response.message
+                }
+              }) {
+                case Success(value) =>
+                  value match {
+                    case tuple @ (status: Int, message: String) => complete(status, message)
+                  }
```
myazinn, line 46: Sorry, I didn't notice this first time. If you want to disassembly a tuple, you can do it in one place
```
case Success((status, message)) =>
```
No need for nested pattern matching here :)

manvicha: Thanks, the code lines has been reduced. The change was applied

PR #72, opened 2020-05-22
Epmlstrcmw 153 typed http client

OstapStrashevskii:
Parametrize HTTP Client, adding new Types for Git Lab API, refactor GitLabVersioning

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
@@ -1,69 +1,143 @@
 package cromwell.pipeline.controller
 
+/**
+ * A class to provide REST operations using PlayJson
+ * @get,
+ * @post,
+ * @put methods have Response generic type, which is
+ * @ResponseBoby type of its Response
+ *
+ * Both ResponseBody (B) and Payload (P) are parametrized
+ * To use this client you have to have implicit Reads or Writes for your entities in your DTO
+ *
+ * Response processed in to steps:
+ * 1) firstly it is filtered by Akka success/failure HttpResponse type using response.status.isSuccess() in @responsify method
+ * 2) secondly it is filtered if not valid in @validateResponse method
+ * as a result if you have both success statusCode and it is validated successfully, you will get [[cromwell.pipeline.service.SuccessResponseBody]],
+ * otherwise [[cromwell.pipeline.service.FailureResponseBody]]
+ *
+ *
+ * @version 1.0
+ */
 import akka.actor.ActorSystem
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.model.{ HttpMethods, HttpRequest, HttpResponse, Uri }
+import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.stream.ActorMaterializer
-import cromwell.pipeline.service.{ HttpClient, Response }
+import cromwell.pipeline.service.{ FailureResponseBody, HttpClient, Response, SuccessResponseBody }
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+import play.api.libs.json.{ Json, Reads, Writes }
 
 import scala.concurrent.duration._
 import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success, Try }
 
+/**
+ * @constructor Create a new client with a `actorSystem` and `materializer`.
+ * @param actorSystem
+ * @param materializer
+ */
 class AkkaHttpClient(implicit actorSystem: ActorSystem, materializer: ActorMaterializer) extends HttpClient {
   private val expirationTime: FiniteDuration = 300.millis
-  override def get(url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
-    implicit ec: ExecutionContext
-  ): Future[Response] = {
+
+  override def get[B](url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
+    implicit ec: ExecutionContext,
+    f: Reads[B]
+  ): Future[Response[B]] = {
     val futureResponse: Future[HttpResponse] = Http().singleRequest(
       HttpRequest(method = HttpMethods.GET, uri = Uri(url).withQuery(Query(params))).withHeaders(parseHeaders(headers))
     )
     responsify(futureResponse)
   }
 
-  private def responsify(futureResponse: Future[HttpResponse])(implicit ec: ExecutionContext): Future[Response] =
+  override def post[B, P](
+    url: String,
+    params: Map[String, String] = Map(),
+    headers: Map[String, String] = Map(),
+    payload: P
+  )(implicit ec: ExecutionContext, bf: Reads[B], pf: Writes[P]): Future[Response[B]] = {
+    val futureResponse: Future[HttpResponse] = Http().singleRequest(
+      HttpRequest(method = HttpMethods.POST, uri = Uri(url).withQuery(Query(params)))
+        .withEntity(Json.stringify(Json.toJson(payload)))
+        .withHeaders(parseHeaders(headers))
+    )
+    responsify[B](futureResponse)
+  }
+
+  override def put[B, P](
+    url: String,
+    params: Map[String, String] = Map(),
+    headers: Map[String, String] = Map(),
+    payload: P
+  )(implicit ec: ExecutionContext, bf: Reads[B], pf: Writes[P]): Future[Response[B]] = {
+    val futureResponse: Future[HttpResponse] = Http().singleRequest(
+      HttpRequest(method = HttpMethods.PUT, uri = Uri(url).withQuery(Query(params)))
+        .withEntity(Json.stringify(Json.toJson(payload)))
+        .withHeaders(parseHeaders(headers))
+    )
+    responsify(futureResponse)
+  }
+
+  /**
+   * Filters unsuccessful responses
+   *
+   * @return FailureResponseBody if [[akka.http.scaladsl.model.StatusCode]] is not succeed, go to validation otherwise
+   */
+  private def responsify[B](
+    futureResponse: Future[HttpResponse]
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
     futureResponse.flatMap { response =>
-      response.entity.toStrict(expirationTime).map { body =>
-        Response(
-          response.status.intValue(),
-          body.data.utf8String,
-          response.headers.foldLeft(Map[String, List[String]]()) {
-            case (acc, header) =>
-              acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
-          }
-        )
+      if (response.status.isSuccess()) {
+        validateResponse(response)
+      } else {
+        failureResponse(response)
       }
+    }
+
+  /**
+   * Parse and validate response body
+   *
+   * @return SuccessResponseBody(value) if body json is valid or FailureResponseBody otherwise
+   */
+  private def validateResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
+    Unmarshal(response.entity).to[B].transform {
+      case Success(body) =>
+        Try(
+          Response[B](
+            response.status.intValue(),
+            SuccessResponseBody(body),
+            parseHeaders(response)
+          )
+        )
+      case Failure(exception) =>
+        Try(
+          Response[B](
+            response.status.intValue(),
+            FailureResponseBody(s"Could not parse and validate response. (errors: ${exception.getMessage})"),
+            parseHeaders(response)
+          )
+        )
+    }
+
+  private def failureResponse[B](response: HttpResponse)(implicit ec: ExecutionContext): Future[Response[B]] =
+    response.entity.toStrict(expirationTime).map { body =>
+      Response[B](
+        response.status.intValue(),
+        FailureResponseBody(body.toString()),
+        parseHeaders(response)
+      )
+    }
+
+  private def parseHeaders(response: HttpResponse): Map[String, Seq[String]] =
+    response.headers.foldLeft(Map[String, List[String]]()) {
+      case (acc, header) =>
+        acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
     }
 
   private def parseHeaders(headers: Map[String, String]): List[RawHeader] =
     headers.map { case (param, value) => RawHeader(param, value) }.toList
-
-  override def post(
-    url: String,
-    params: Map[String, String] = Map(),
-    headers: Map[String, String] = Map(),
-    payload: String
-  )(implicit ec: ExecutionContext): Future[Response] = {
-    val futureResponse: Future[HttpResponse] = Http().singleRequest(
-      HttpRequest(method = HttpMethods.POST, uri = Uri(url).withQuery(Query(params)))
-        .withEntity(payload)
-        .withHeaders(parseHeaders(headers))
-    )
-    responsify(futureResponse)
-  }
-
-  override def put(
-    url: String,
-    params: Map[String, String] = Map(),
-    headers: Map[String, String] = Map(),
-    payload: String
-  )(implicit ec: ExecutionContext): Future[Response] = {
-    val futureResponse: Future[HttpResponse] = Http().singleRequest(
-      HttpRequest(method = HttpMethods.PUT, uri = Uri(url).withQuery(Query(params)))
-        .withEntity(payload)
-        .withHeaders(parseHeaders(headers))
-    )
-    responsify(futureResponse)
-  }
 }
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -3,23 +3,34 @@ package cromwell.pipeline.controller
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
+import cromwell.pipeline.datastorage.dto.SuccessResponseMessage
 import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
-import cromwell.pipeline.datastorage.dto.{FileContent, ProjectBuildConfigurationRequest, ProjectUpdateFileRequest}
-import cromwell.pipeline.service.VersioningException.{FileException, GitException, HttpException, ProjectException, RepositoryException}
-import cromwell.pipeline.service.{ProjectFileService, VersioningException}
+import cromwell.pipeline.service.VersioningException.{
+  FileException,
+  GitException,
+  HttpException,
+  ProjectException,
+  RepositoryException
+}
+import cromwell.pipeline.datastorage.dto.{
+  ProjectBuildConfigurationRequest,
+  ProjectFileContent,
+  ProjectUpdateFileRequest
+}
+
+import cromwell.pipeline.service.{ ProjectFileService }
 import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
 
-import scala.concurrent.ExecutionContext
-import scala.util.{Failure, Success}
+import scala.concurrent.{ ExecutionContext }
+import scala.util.{ Failure, Success }
 
-class ProjectFileController(wdlService: ProjectFileService)(
-  implicit val executionContext: ExecutionContext
-) {
+class ProjectFileController(wdlService: ProjectFileService)(implicit val executionContext: ExecutionContext) {
+
   val route: AccessTokenContent => Route = _ =>
     concat(
       path("files" / "validation") {
         post {
-          entity(as[FileContent]) { request =>
+          entity(as[ProjectFileContent]) { request =>
             onComplete(wdlService.validateFile(request)) {
               case Success(Left(e)) => complete(StatusCodes.Conflict, e.errors)
               case Success(_)       => complete(StatusCodes.OK)
@@ -33,22 +44,25 @@ class ProjectFileController(wdlService: ProjectFileService)(
           entity(as[ProjectUpdateFileRequest]) {
             request =>
               onComplete(for {
-                validateResponse <- wdlService.validateFile(FileContent(request.projectFile.content))
+                validateResponse <- wdlService.validateFile(request.projectFile.content)
                 uploadResponse <- wdlService.uploadFile(request.project, request.projectFile, request.version)
               } yield {
                 (validateResponse, uploadResponse) match {
-                  case (Right(_), Right(response)) => StatusCodes.OK.intValue -> response
-                  case (Left(_), Right(response))  => StatusCodes.Created.intValue -> response
-                  case (_, Left(response))         => response match {
-                    case HttpException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                    case FileException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                    case GitException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                    case RepositoryException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                    case ProjectException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                  }
+                  case (Right(_), Right(responseMessage)) => StatusCodes.OK.intValue -> responseMessage
+                  case (Left(_), Right(responseMessage))  => StatusCodes.Created.intValue -> responseMessage
+                  case (_, Left(response)) =>
+                    response match {
+                      case HttpException(message)       => StatusCodes.UnprocessableEntity.intValue -> message
+                      case FileException(message)       => StatusCodes.UnprocessableEntity.intValue -> message
+                      case GitException(message)        => StatusCodes.UnprocessableEntity.intValue -> message
+                      case RepositoryException(message) => StatusCodes.UnprocessableEntity.intValue -> message
+                      case ProjectException(message)    => StatusCodes.UnprocessableEntity.intValue -> message
+                    }
                 }
               }) {
-                case Success((status, message)) => complete(status, message)
+                case Success((status, p @ SuccessResponseMessage(_))) =>
+                  complete((status, p))
+                case Success((status, message)) => complete((status, s"File have not uploaded due to $message"))
                 case Failure(e)                 => complete(StatusCodes.InternalServerError, e.getMessage)
               }
           }
--- a/controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
@@ -1,23 +1,28 @@
 package cromwell.pipeline.controller
 
-import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
-import akka.stream.ActorMaterializer
 import com.github.tomakehurst.wiremock.WireMockServer
 import com.github.tomakehurst.wiremock.client.WireMock._
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration._
-import org.scalatest.{ AsyncWordSpec, BeforeAndAfterAll, Matchers }
+import cromwell.pipeline.service.SuccessResponseBody
+import cromwell.pipeline.utils.{ AkkaTestSources, DummyObject }
+import org.scalatest.{ AsyncWordSpec, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
+import play.api.libs.json.Json
 
-import scala.concurrent.ExecutionContext
-
-class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with BeforeAndAfterAll {
-  implicit val actorSystem: ActorSystem = ActorSystem()
-  implicit val materializer: ActorMaterializer = ActorMaterializer()
-  implicit val ec: ExecutionContext = actorSystem.dispatcher
+class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with AkkaTestSources {
 
   val wireMockServer = new WireMockServer(wireMockConfig().dynamicPort())
   val client = new AkkaHttpClient()
+
+  val dummyObject: DummyObject = DummyObject(12345, "something as object content")
+  val stringifiedRespBody = """{"someId":12345,"someString":"something as object content"}"""
+  val successRespBody = SuccessResponseBody(dummyObject)
+  val applicationJson = "application/json"
+
+  val params = Map("id" -> "1")
+  val headers = Map("Language" -> "eng")
+  val payload = dummyObject
 
   override def beforeAll: Unit = {
     super.beforeAll()
@@ -33,129 +38,132 @@ class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar w
 
     "get" should {
       "return OK response status" taggedAs Controller in {
-        val response = aResponse().withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(get(urlEqualTo("/get?id=1")).willReturn(response))
-
         val get_url = s"${wireMockServer.baseUrl()}/get"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
 
-        client.get(get_url, params, headers).flatMap(_.status shouldBe StatusCodes.OK.intValue)
+        client.get[DummyObject](get_url, params, headers).flatMap(_.status shouldBe StatusCodes.OK.intValue)
       }
 
       "return response with body" taggedAs Controller in {
-        val response = aResponse().withBody("Value").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(get(urlEqualTo("/get?id=1")).willReturn(response))
-
         val get_url = s"${wireMockServer.baseUrl()}/get"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
 
-        client.get(get_url, params, headers).flatMap(_.body shouldBe "Value")
+        client.get[DummyObject](get_url, params, headers).flatMap(_.body shouldBe successRespBody)
       }
 
       "return response with headers" taggedAs Controller in {
-        val response = aResponse().withHeader("TestKey", "TestValue").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("TestKey", "TestValue")
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(get(urlEqualTo("/get?id=1")).willReturn(response))
-
         val get_url = s"${wireMockServer.baseUrl()}/get"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
 
         client
-          .get(get_url, params, headers)
+          .get[DummyObject](get_url, params, headers)
           .flatMap(_.headers.getOrElse("TestKey", List("failed")) shouldBe List("TestValue"))
       }
     }
 
     "post" should {
       "return OK response status" taggedAs Controller in {
-        val response = aResponse().withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val post_url = s"${wireMockServer.baseUrl()}/post"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
-        client.post(post_url, params, headers, payload).flatMap(_.status shouldBe StatusCodes.OK.intValue)
+        client
+          .post[DummyObject, DummyObject](post_url, params, headers, payload)
+          .flatMap(_.status shouldBe StatusCodes.OK.intValue)
       }
 
       "return response with body" taggedAs Controller in {
-        val response = aResponse().withBody("Value").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val post_url = s"${wireMockServer.baseUrl()}/post"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
-        client.post(post_url, params, headers, payload).flatMap(_.body shouldBe "Value")
+        client
+          .post[DummyObject, DummyObject](post_url, params, headers, payload)
+          .flatMap(_.body shouldBe successRespBody)
       }
 
       "return response with headers" taggedAs Controller in {
-        val response = aResponse().withHeader("TestKey", "TestValue").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("TestKey", "TestValue")
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          post(urlEqualTo("/post?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val post_url = s"${wireMockServer.baseUrl()}/post"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
         client
-          .post(post_url, params, headers, payload)
+          .post[DummyObject, DummyObject](post_url, params, headers, payload)
           .flatMap(_.headers.getOrElse("TestKey", List("failed")) shouldBe List("TestValue"))
       }
     }
 
     "put" should {
       "return OK response status" taggedAs Controller in {
-        val response = aResponse().withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val put_url = s"${wireMockServer.baseUrl()}/put"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
-        client.put(put_url, params, headers, payload).flatMap(_.status shouldBe StatusCodes.OK.intValue)
+        client
+          .put[DummyObject, DummyObject](put_url, params, headers, payload)
+          .flatMap(_.status shouldBe StatusCodes.OK.intValue)
       }
 
       "return response with body" taggedAs Controller in {
-        val response = aResponse().withBody("Value").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val put_url = s"${wireMockServer.baseUrl()}/put"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
-        client.put(put_url, params, headers, payload).flatMap(_.body shouldBe "Value")
+        client.put[DummyObject, DummyObject](put_url, params, headers, payload).flatMap(_.body shouldBe successRespBody)
       }
 
       "return response with headers" taggedAs Controller in {
-        val response = aResponse().withHeader("TestKey", "TestValue").withStatus(200)
+        val response = aResponse()
+          .withBody(Json.stringify(Json.toJson(dummyObject)))
+          .withStatus(200)
+          .withHeader("TestKey", "TestValue")
+          .withHeader("Content-Type", applicationJson)
         wireMockServer.stubFor(
-          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo("test payload")).willReturn(response)
+          put(urlEqualTo("/put?id=1")).withRequestBody(equalTo(stringifiedRespBody)).willReturn(response)
         )
-
         val put_url = s"${wireMockServer.baseUrl()}/put"
-        val params = Map("id" -> "1")
-        val headers = Map("Language" -> "eng")
-        val payload = "test payload"
 
         client
-          .put(put_url, params, headers, payload)
+          .put[DummyObject, DummyObject](put_url, params, headers, payload)
           .flatMap(_.headers.getOrElse("TestKey", List("failed")) shouldBe List("TestValue"))
       }
     }
--- a/controllers/src/test/scala/cromwell/pipeline/controller/ProjectFileControllerTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/ProjectFileControllerTest.scala
@@ -23,7 +23,7 @@ class ProjectFileControllerTest extends AsyncWordSpec with Matchers with Scalate
     val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId)
 
     "validate file" should {
-      val content = FileContent("task hello {}")
+      val content = ProjectFileContent("task hello {}")
 
       "return OK response to valid file" taggedAs Controller in {
         when(projectFileService.validateFile(content)).thenReturn(Future.successful(Right(())))
@@ -43,27 +43,29 @@ class ProjectFileControllerTest extends AsyncWordSpec with Matchers with Scalate
     }
 
     "upload file" should {
+      import cromwell.pipeline.datastorage.dto.SuccessResponseMessage
+
       val version = PipelineVersion("v0.0.2")
       val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId)
       val project = TestProjectUtils.getDummyProject()
-      val projectFile = ProjectFile(Paths.get("folder/test.txt"), "file context")
+      val projectFileContent = ProjectFileContent("file context")
+      val projectFile = ProjectFile(Paths.get("folder/test.txt"), projectFileContent)
       val request = ProjectUpdateFileRequest(project, projectFile, Some(version))
-      val content = FileContent(projectFile.content)
 
       "return OK response for valid request with a valid file" taggedAs Controller in {
-        when(projectFileService.validateFile(content)).thenReturn(Future.successful(Right(())))
+        when(projectFileService.validateFile(projectFileContent)).thenReturn(Future.successful(Right(())))
         when(projectFileService.uploadFile(project, projectFile, Some(version)))
-          .thenReturn(Future.successful(Right("Success")))
+          .thenReturn(Future.successful(Right(SuccessResponseMessage("Success"))))
         Post("/files", request) ~> projectFileController.route(accessToken) ~> check {
           status shouldBe StatusCodes.OK
         }
       }
 
       "return Precondition File response for valid request with an invalid file" taggedAs Controller in {
-        when(projectFileService.validateFile(content))
+        when(projectFileService.validateFile(projectFileContent))
           .thenReturn(Future.successful(Left(ValidationError(List("Miss close bracket")))))
         when(projectFileService.uploadFile(project, projectFile, Some(version)))
-          .thenReturn(Future.successful(Right("Success")))
+          .thenReturn(Future.successful(Right(SuccessResponseMessage("Success"))))
         Post("/files", request) ~> projectFileController.route(accessToken) ~> check {
           status shouldBe StatusCodes.Created
         }
@@ -74,14 +76,14 @@ class ProjectFileControllerTest extends AsyncWordSpec with Matchers with Scalate
           .thenReturn(Future.successful(Left(VersioningException.HttpException("Bad request"))))
         Post("/files", request) ~> projectFileController.route(accessToken) ~> check {
           status shouldBe StatusCodes.UnprocessableEntity
-          entityAs[String] shouldBe "Bad request"
+          entityAs[String] shouldBe "File have not uploaded due to Bad request"
         }
       }
     }
 
     "build configuration" should {
       val projectId = TestProjectUtils.getDummyProjectId
-      val projectFile = ProjectFile(Paths.get("/home/test/file"), "{some context}")
+      val projectFile = ProjectFile(Paths.get("/home/test/file"), ProjectFileContent("{some context}"))
       val configuration = ProjectConfiguration(
         projectId,
         List(
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -49,7 +49,7 @@ object Dependencies {
   val akkaTestKit = "com.typesafe.akka" %% "akka-testkit" % Version.akka % "test,it"
   val akkaHttpTestKit = "com.typesafe.akka" %% "akka-http-testkit" % Version.akkaHttp % "test,it"
   val scalaCheck = "org.scalacheck" %% "scalacheck" % Version.scalaCheck % "test,it"
-  val scalaTest = "org.scalatest" %% "scalatest" % Version.scalaTest % "test,it"
+  val scalaTest = "org.scalatest" %% "scalatest" % Version.scalaTest % "test, it"
   val mockito = "org.mockito" % "mockito-all" % Version.mockito % Test
   val tcCore = "com.dimafeng" %% "testcontainers-scala" % Version.tcCore % Test
   val tcPostgres = "org.testcontainers" % "postgresql" % Version.tcPostgres % Test
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/File.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/File.scala
@@ -1,16 +1,16 @@
 package cromwell.pipeline.datastorage.dto
 
 import play.api.libs.functional.syntax._
-import play.api.libs.json.{ Format, JsPath }
+import play.api.libs.json.{ JsPath, OFormat }
 
 object File {
-  case class UpdateFileRequest(content: String, commitMessage: String, branch: String)
+  case class UpdateFileRequest(content: ProjectFileContent, commitMessage: String, branch: String)
 
   object UpdateFileRequest {
-    implicit val updateFileRequestFormat: Format[UpdateFileRequest] =
-      (JsPath \ "branch")
-        .format[String]
-        .and((JsPath \ "content").format[String])
-        .and((JsPath \ "commit_message").format[String])(UpdateFileRequest.apply, unlift(UpdateFileRequest.unapply))
+    implicit val updateFileRequestFormat: OFormat[UpdateFileRequest] =
+      (JsPath \ "content")
+        .format[ProjectFileContent]
+        .and((JsPath \ "commit_message").format[String])
+        .and((JsPath \ "branch").format[String])(UpdateFileRequest.apply, unlift(UpdateFileRequest.unapply))
   }
 }
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
@@ -10,13 +10,13 @@ import play.api.libs.json._
 import slick.lifted.MappedTo
 
 final case class Project(
-                          projectId: ProjectId,
-                          ownerId: UserId,
-                          name: String,
-                          active: Boolean,
-                          repository: Option[Repository] = None,
-                          visibility: Visibility = Private
-                        ) {
+  projectId: ProjectId,
+  ownerId: UserId,
+  name: String,
+  active: Boolean,
+  repository: Option[Repository] = None,
+  visibility: Visibility = Private
+) {
   def withRepository(repositoryPath: Option[String]): Project =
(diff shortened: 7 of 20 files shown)


controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,115 @@ import akka.actor.ActorSystem
...
+        response.status.intValue(),
+        NegativeResponseBody(pb.toString()),
+        parseHeaders(response)
+      )
+    }
+
+  private def parseHeaders(response: HttpResponse): Map[String, Seq[String]] =
+    response.headers.foldLeft(Map[String, List[String]]()) {
+      case (acc, header) =>
+        acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
```
AndreiMaksimenko, line 138: As I understand header.value() is Seq (or single element) why you concat it with acc - which is Map?

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,115 @@ import akka.actor.ActorSystem
...
 
+  private def getResponseBody[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[JsResult[B]] =
+    response.entity.toStrict(expirationTime).map { body =>
+      Json.fromJson[B](parseResponseBody(body))
+    }
+
+  private def parseResponseBody(e: HttpEntity.Strict): JsValue = {
+    println("CLASS=" + e.getClass)
```
AndreiMaksimenko: Remove this debug process: println!

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,115 @@ import akka.actor.ActorSystem
...
     headers.map { case (param, value) => RawHeader(param, value) }.toList
 
-  override def post(
+  override def post[B, P](
     url: String,
     params: Map[String, String] = Map(),
     headers: Map[String, String] = Map(),
-    payload: String
-  )(implicit ec: ExecutionContext): Future[Response] = {
+    payload: P
```
AndreiMaksimenko: Do we really need to have a typed payload?

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -16,7 +16,7 @@ class ProjectFileController(wdlService: ProjectFileService)(implicit val executi
     concat(
       path("files" / "validation") {
         post {
-          entity(as[FileContent]) { request =>
+          entity(as[ProjectFileContent]) { request =>
```
AndreiMaksimenko, line 33: Thank you for this rename!

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -33,129 +46,146 @@ class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar w
...
       }
 
       "return response with body" taggedAs Controller in {
-        val response = aResponse().withBody("Value").withStatus(200)
+        val response = aResponse().withBody(Json.stringify(Json.toJson(dummyObject))).withStatus(200)
         wireMockServer.stubFor(get(urlEqualTo("/get?id=1")).willReturn(response))
 
         val get_url = s"${wireMockServer.baseUrl()}/get"
         val params = Map("id" -> "1")
         val headers = Map("Language" -> "eng")
```
AndreiMaksimenko: You can move it to constants, and don't repeat yourself!

controllers/src/test/scala/cromwell/pipeline/controller/ProjectFileControllerTest.scala
```diff
@@ -52,7 +52,8 @@ class ProjectFileControllerTest extends AsyncWordSpec with Matchers with Scalate
       val version = Version("v.0.0.2", "commit message", "this project", Commit("commit_12"))
       val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId.value)
       val project = TestProjectUtils.getDummyProject()
-      val projectFile = ProjectFile(Paths.get("folder/test.txt"), "file context")
+      val projectFileContent = ProjectFileContent("task hello {}")
```
AndreiMaksimenko: We can move it to constants.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -3,38 +3,56 @@ package cromwell.pipeline.service
 import java.net.URLEncoder
 import java.nio.file.Paths
 
+import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
+import akka.stream.ActorMaterializer
 import cromwell.pipeline.datastorage.dao.repository.utils.TestProjectUtils
 import cromwell.pipeline.datastorage.dto.File.UpdateFileRequest
+import cromwell.pipeline.datastorage.dto.{ Commit, File, Message, Project, ProjectFile, Version }
 import cromwell.pipeline.datastorage.dto._
```
AndreiMaksimenko: I think we need just one of these lines

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -3,38 +3,56 @@ package cromwell.pipeline.service
...
+
   import ProjectContext._
   import ProjectFileContext._
 
   "GitLabProjectVersioning" when {
 
     "createRepository" should {
 
-      def payload(project: Project): String = Json.stringify(Json.toJson(project))
+      implicit val formatter: OFormat[Project] = Project.projectFormat
```
AndreiMaksimenko: I think we can just import implicit from project

AndreiMaksimenko (commented):
Good job, please take a look at my comments!

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -3,38 +3,56 @@ package cromwell.pipeline.service
...
 import cromwell.pipeline.datastorage.dto.File.UpdateFileRequest
+import cromwell.pipeline.datastorage.dto.{ Commit, File, Message, Project, ProjectFile, Version }
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.utils.{ ApplicationConfig, GitLabConfig, HttpStatusCodes }
 import org.mockito.Mockito.when
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.{ AsyncWordSpec, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
-import play.api.libs.json.Json
+import play.api.libs.json.{ OFormat }
```
Dmitrii-Shevchenko: brackets abuse

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: truth, done

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: you are right

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: done

controllers/src/test/scala/cromwell/pipeline/controller/ProjectFileControllerTest.scala
OstapStrashevskii: This list IMHO, looks little more clearly, dont it?

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
OstapStrashevskii: that's not mine, but I have done this, good idea, completely agreed

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
OstapStrashevskii: FileContent sounds very abstract, I don't know shall we have something else with it

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: affirmative) I suppose, yes, this was discussed. Maybe it should be discussed additionally

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: uff! thanks, done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: @GrigoriiBerezin can we leak something while converting Seq to Map? I think we can we lose some headers info, Is not it?

GrigoriiBerezin: The thing is that we can have a few headers with the same name and if we want to save them all into the map, Ostap did it right

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
GrigoriiBerezin: Payload should definitely be typed

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[Response[B]] =
     futureResponse.flatMap { response =>
-      response.entity.toStrict(expirationTime).map { body =>
-        Response(
+      response.status match {
+        case StatusCodes.Created    => positiveResponse(response)(ec, f)
+        case StatusCodes.Accepted   => positiveResponse(response)(ec, f)
+        case StatusCodes.BadRequest => negativeResponse(response)(ec)
+        case StatusCodes.NotFound   => negativeResponse(response)(ec)
+        case _                      => negativeResponse(response)(ec)
```
GrigoriiBerezin: You don't need to pass implicit parameters to the methods

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+      }
+    }
+
+  private def positiveResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[Response[B]] =
+    if (response.headers.contains(httpContentTypeJson)) {
+      getResponseBody[B](response)(ec, f).map { responseBody =>
```
GrigoriiBerezin: Same here

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+
+  private def parseHeaders(response: HttpResponse): Map[String, Seq[String]] =
+    response.headers.foldLeft(Map[String, List[String]]()) {
+      case (acc, header) =>
+        acc + (header.name() -> (header.value() :: acc.getOrElse(header.name(), Nil)))
+    }
+
+  private def getResponseBody[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[JsResult[B]] =
```
GrigoriiBerezin: For this method Reads instead of Format is enough to work correctly

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
 class AkkaHttpClient(implicit actorSystem: ActorSystem, materializer: ActorMaterializer) extends HttpClient {
   private val expirationTime: FiniteDuration = 300.millis
+  private val httpContentTypeJson = "application/json"
 
-  override def get(url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
-    implicit ec: ExecutionContext
-  ): Future[Response] = {
+  override def get[B](url: String, params: Map[String, String] = Map(), headers: Map[String, String] = Map())(
+    implicit ec: ExecutionContext,
+    f: Format[B]
```
GrigoriiBerezin: Reads is enough

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
     val futureResponse: Future[HttpResponse] = Http().singleRequest(
       HttpRequest(method = HttpMethods.GET, uri = Uri(url).withQuery(Query(params))).withHeaders(parseHeaders(headers))
     )
     responsify(futureResponse)
   }
 
-  private def responsify(futureResponse: Future[HttpResponse])(implicit ec: ExecutionContext): Future[Response] =
+  private def responsify[B](
+    futureResponse: Future[HttpResponse]
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[Response[B]] =
```
GrigoriiBerezin: Reads is enough

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+        case StatusCodes.Accepted   => positiveResponse(response)(ec, f)
+        case StatusCodes.BadRequest => negativeResponse(response)(ec)
+        case StatusCodes.NotFound   => negativeResponse(response)(ec)
+        case _                      => negativeResponse(response)(ec)
+      }
+    }
+
+  private def positiveResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Format[B]): Future[Response[B]] =
```
GrigoriiBerezin: Reads is enough

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
 
-  override def post(
+  override def post[B, P](
     url: String,
     params: Map[String, String] = Map(),
     headers: Map[String, String] = Map(),
-    payload: String
-  )(implicit ec: ExecutionContext): Future[Response] = {
+    payload: P
+  )(implicit ec: ExecutionContext, bf: Format[B], pf: Format[P]): Future[Response[B]] = {
```
GrigoriiBerezin: Reads for bf and Writes for pf

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
 
-  override def put(
+  override def put[B, P](
     url: String,
     params: Map[String, String] = Map(),
     headers: Map[String, String] = Map(),
-    payload: String
-  )(implicit ec: ExecutionContext): Future[Response] = {
+    payload: P
+  )(implicit ec: ExecutionContext, bf: Format[B], pf: Format[P]): Future[Response[B]] = {
```
GrigoriiBerezin: Reads for bf and Writes for pf

GrigoriiBerezin (changes requested):
Nicely done, great work. But there is a few moment to discuss/change/argue

Dmitrii-Shevchenko (approved):
Good work!

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: that's right, I suppose. Done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done, to change it with more concrete parser implementation is a very good idea, it is easier to separate respbody and payload, so thanks

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: nice, done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: nice, done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done

GrigoriiBerezin (approved):
Great work!

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
myazinn: @AndreiMaksimenko @GrigoriiBerezin 
Don't forget to resolve your conversations if they are no longer relevant.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
     futureResponse.flatMap { response =>
-      response.entity.toStrict(expirationTime).map { body =>
-        Response(
+      response.status match {
+        case StatusCodes.Created    => positiveResponse(response)
+        case StatusCodes.Accepted   => positiveResponse(response)
+        case StatusCodes.BadRequest => negativeResponse(response)
+        case StatusCodes.NotFound   => negativeResponse(response)
+        case _                      => negativeResponse(response)
```
myazinn: I don't think there is a need to explicitly pattern match over so many cases if we handle all of them in the same way.
```
case StatusCodes.Created | StatusCodes.Accepted => positive
case _ => negative
```

AndreiMaksimenko (approved):

IaroslavTavchenkov (approved):
Good job! :)

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,113 @@ import akka.actor.ActorSystem
...
+      }
+    }
+
+  private def positiveResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
+    if (response.headers.contains(httpContentTypeJson)) {
```
myazinn: Why are you not using `Unmarshal` here?

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -6,11 +6,19 @@ import akka.stream.ActorMaterializer
...
 import com.github.tomakehurst.wiremock.client.WireMock._
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration._
+import cromwell.pipeline.service.NegativeResponseBody
 import org.scalatest.{ AsyncWordSpec, BeforeAndAfterAll, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
+import play.api.libs.json.{ Json, OFormat }
 
 import scala.concurrent.ExecutionContext
 
+case class DummyObject(someId: Int, someString: String)
```
myazinn: Move this class to some utils classes

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -6,11 +6,19 @@ import akka.stream.ActorMaterializer
...
+
+object DummyObject {
+  implicit lazy val DummyObjectFormat: OFormat[DummyObject] = Json.format[DummyObject]
```
myazinn:
```suggestion
  implicit lazy val dummyObjectFormat: OFormat[DummyObject] = Json.format[DummyObject]
```

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/File.scala
```diff
@@ -1,16 +1,18 @@
...
-    implicit val updateFileRequestFormat: Format[UpdateFileRequest] =
+    implicit val updateFileRequestFormat: OFormat[UpdateFileRequest] =
       (JsPath \ "branch")
         .format[String]
-        .and((JsPath \ "content").format[String])
+        .and((JsPath \ "content").format[ProjectFileContent])
         .and((JsPath \ "commit_message").format[String])(UpdateFileRequest.apply, unlift(UpdateFileRequest.unapply))
   }
+
+  object GetFileRequest
```
myazinn: What is this?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -65,26 +65,36 @@ object ProjectUpdateRequest {
 final case class Version(name: String, message: String, target: String, commit: Commit)
 object Version {
   implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
+  implicit val seqVersionPlayFormat: Format[Seq[Version]] =
+    Format(Reads.seq(versionPlayFormat), Writes.seq(versionPlayFormat))
```
myazinn: `play-json` can format sequences automatically.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -21,44 +20,57 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
             )
             .map {
-              case Response(HttpStatusCodes.OK, _, _) => Right("Create new file")
-              case Response(_, body, _)               => Left(VersioningException(body))
+              case Response(HttpStatusCodes.OK, _, _)         => Right("Create new file")
+              case Response(_, NegativeResponseBody(body), _) => Left(VersioningException(body))
+              case _ =>
+                Left(
+                  VersioningException(
+                    "status is not Ok, but response is not negative, check status and possible positive responses"
```
myazinn: The existence of this line indicates that something went wrong during the implementation of the HTTP client.
Separating bad and good responses is a good idea. But you must make sure that whoever uses your client won't have to assume that something went wrong in the HTTP client implementation.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -21,44 +20,57 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
   }
 
   override def updateFiles(project: Project, projectFiles: ProjectFiles)(
     implicit ec: ExecutionContext
   ): AsyncResult[List[String]] = ???
 
-  override def createRepository(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] =
+  override def createRepository(project: Project)(implicit ec: ExecutionContext): AsyncResult[Project] = {
+
+    import cromwell.pipeline.datastorage.dto.Project._
```
myazinn: Move it to other imports at the top of the file.

services/src/main/scala/cromwell/pipeline/service/HttpClient.scala
```diff
@@ -1,17 +1,30 @@
...
+    pf: Writes[P]
+  ): Future[Response[B]]
 }
 
-case class Response(status: Int, body: String, headers: Map[String, Seq[String]])
+case class Response[B](status: Int, body: ResponseBody[B], headers: Map[String, Seq[String]])
+
+sealed trait ResponseBody[B]
+
+final case class PositiveResponseBody[B](bodyObject: B) extends ResponseBody[B]
```
myazinn: Let's stick to scala naming conventions and use something like `success` and `failure` instead of `positive` and `negative`.

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -12,29 +14,42 @@ import org.mockito.Mockito.when
...
 import org.scalatest.{ AsyncWordSpec, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
-import play.api.libs.json.Json
+import play.api.libs.json.OFormat
 
-import scala.concurrent.Future
+import scala.concurrent.{ ExecutionContext, Future }
 
 class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with Matchers with MockitoSugar {
+  implicit val actorSystem: ActorSystem = ActorSystem()
```
myazinn: Creating your own actor system in test is a bad idea. Especially since you are not even closing it.

myazinn (changes requested):
Great job!
But nothing is perfect :)

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: of course, but when I use generic formatter, I need to set formatters explicitly(from generic parameter), that is why I need to use this formatter. This dilemma is in generic method which need type as a parameter and deduce formatter from the type, and method's return type. I have to choose either an accurate parameter type or an accurate return type. Maybe it's too complicated to understand this as text, It should be easier to discuss it verbally.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/File.scala
OstapStrashevskii: its a shame, I forgot this artefact, thanks you are an eagle eye)

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
OstapStrashevskii: done

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
OstapStrashevskii: done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: while I am working with headers and not only body, I need

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: As I remember this was made just in order to make the match exhaustive. For the cases where we try to make a request and get unexpected success response(which have not already exist in httpclient match)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: done

services/src/main/scala/cromwell/pipeline/service/HttpClient.scala
OstapStrashevskii: done

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: What is the best practice in this case?

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: used, refactored

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: done it with class in utils module

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,100 @@ import akka.actor.ActorSystem
...
+      Response[B](
+        response.status.intValue(),
+        SuccessResponseBody(b),
+        parseHeaders(response)
+      )
+    }
+
+  private def failureResponse[B](response: HttpResponse)(implicit ec: ExecutionContext): Future[Response[B]] =
+    response.entity.toStrict(expirationTime).map { body =>
+      val pb = parseResponseBody(body)
```
myazinn, line 54: I think we cannot assume that the failed response will return a `json`.  Therefore, you should not parse this body to a `JsValue`.

myazinn: I think we cannot assume that the failed response will return a `json`. Therefore, you should not parse this body to a `JsValue`.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,100 @@ import akka.actor.ActorSystem
...
       HttpRequest(method = HttpMethods.PUT, uri = Uri(url).withQuery(Query(params)))
-        .withEntity(payload)
+        .withEntity(Json.stringify(Json.toJson(payload)))
         .withHeaders(parseHeaders(headers))
     )
     responsify(futureResponse)
   }
 }
+
+case class ParseJsonException(message: String) extends Exception(message)
```
myazinn: Unused class

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
```diff
@@ -2,23 +2,31 @@ package cromwell.pipeline.controller
...
-  implicit val materializer: ActorMaterializer = ActorMaterializer()
-  implicit val ec: ExecutionContext = actorSystem.dispatcher
+class AkkaHttpClientTest extends AsyncWordSpec with Matchers with MockitoSugar with AkkaTestSources {
 
+  implicit lazy val system: ActorSystem = actorSystem
   val wireMockServer = new WireMockServer(wireMockConfig().dynamicPort())
   val client = new AkkaHttpClient()
 
+  val dummyObject: DummyObject = DummyObject(12345, "something as object content")
+  val stringifiedRespBody = "{\"someId\":12345,\"someString\":\"something as object content\"}"
```
myazinn: Use `"""` string instead of `"` to avoid escape characters. It will make the code more readable.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
myazinn: We've discussed it and you said that you'll delete it. But you didn't.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -65,26 +65,36 @@ object ProjectUpdateRequest {
...
 object ProjectFile {
-  implicit object ProjectFileFormat extends Format[ProjectFile] {
-    override def reads(json: JsValue): JsResult[ProjectFile] =
-      JsSuccess(ProjectFile(Paths.get((json \ "path").as[String]), (json \ "content").as[String]))
 
-    override def writes(o: ProjectFile): JsValue = JsObject(
-      Seq("path" -> JsString(o.path.toString), "content" -> JsString(o.content))
-    )
+  implicit object pathFormat extends Format[Path] {
+    override def writes(o: Path): JsValue = JsObject(Seq("path" -> JsString(o.toString)))
```
myazinn:
```suggestion
    override def writes(o: Path): JsValue = JsString(o.toString)
```
Don't forget to change reads in a similar way

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,67 +4,100 @@ import akka.actor.ActorSystem
...
-        .withEntity(payload)
+        .withEntity(Json.stringify(Json.toJson(payload)))
         .withHeaders(parseHeaders(headers))
     )
-    responsify(futureResponse)
+    responsify[B](futureResponse)
   }
 
-  override def put(
+  override def put[B, P](
```
myazinn: In my opinion, it is more convenient to use the first type parameter as an input and a second type parameter as an output.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -125,3 +129,8 @@ object ProjectUpdateFileRequest {
       unlift(ProjectUpdateFileRequest.unapply)
     )
 }
+
+final case class Message(text: String)
```
myazinn: It took some time to understand why you need such a class.
You are using this class in a large number of places. Are you sure that all of these places will return some `json` that contains the `text` field?
And you must give it a more meaningful name.

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
```diff
@@ -0,0 +1,22 @@
...
+  implicit lazy val materializer: ActorMaterializer = ActorMaterializer()(actorSystem)
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+    actorSystem
+    materializer
+  }
+
+  override protected def afterAll(): Unit = {
+    super.afterAll()
```
myazinn: `super.afterAll` must be called after your custom finalizing actions.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -71,19 +81,25 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
-          if (resp.status != HttpStatusCodes.OK)
-            Left(VersioningException(s"Could not take versions. Response status: ${resp.status}"))
-          else {
-            val parsedVersions: JsResult[Seq[Version]] = Json.parse(resp.body).validate[Seq[Version]]
-            parsedVersions match {
-              case JsSuccess(value, _) => Right(value)
-              case JsError(errors)     => Left(VersioningException(s"Could not parse GitLab response. (errors: $errors)"))
-            }
+          resp.status match {
+            case HttpStatusCodes.OK =>
```
myazinn: I think there is no need to pattern match on response status. Pattern matching on the response body should be enough.
However, if you think that you want two pattern matches on both of these things, do not use nested pattern matching. You did it right on lines 40-42.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -107,20 +123,30 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
         s"${config.url}/projects/${project.repository}/repository/files/$filePath/raw",
         Map("ref" -> fileVersion),
         config.token
       )
       .map(
         resp =>
           resp.status match {
-            case HttpStatusCodes.OK => Right(ProjectFile(path, resp.body))
-            case _                  => Left(VersioningException(s"Exception. Response status: ${resp.status}"))
+            case HttpStatusCodes.OK =>
```
myazinn: Same problem with nested matches.

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
```diff
@@ -0,0 +1,22 @@
+package cromwell.pipeline.utils
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import org.scalatest.{ BeforeAndAfterAll, Suite }
+
+trait AkkaTestSources extends BeforeAndAfterAll {
+  this: Suite =>
+  lazy val actorSystem = ActorSystem("test")
```
myazinn: I think an `ActorSystem` can be marked as an `implicit` too.
In this case, you won't have to do weird things like
`implicit val system = actorSystem`
as you did somewhere above.

myazinn (changes requested):
Your PR is noticeably improved! Yet there is some work to be done.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: replaced

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: removed

controllers/src/test/scala/cromwell/pipeline/controller/AkkaHttpClientTest.scala
OstapStrashevskii: done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: have done as follows
```
    response.entity.toStrict(expirationTime).map { body =>
      Response[B](
        response.status.intValue(),
        FailureResponseBody(body.toString()),
        parseHeaders(response)
      )
    }
```
Should I use just toString here?

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
OstapStrashevskii: moved

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
OstapStrashevskii: did it

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: **My expectations about it:**
This, I suppose, will be replaced later by classes with more concrete replies. Now it describes common cases: For example, it needs for cases where GitLab just replies to us with a notification that it created file or not, etc
like:
```
{
 'smth':'project created'
}
```

For responses which are not bear some informative work, have any informative value.(we have a response status, and we know that it succeed). Or for situations where information in response bodies will be evaluated later...

So the better name for it may be successResponseMessage ?

**And "pragmatic" answer:**
The point is that @GrigoriiBerezin has some "mock" answers in his test cases set, and I have parametrized client, so I had to encapsulate String response bodies into types.

I have  ADT:
```
final case class SuccessResponseBody[B](bodyObject: B) extends ResponseBody[B]
final case class FailureResponseBody[B](error: String) extends ResponseBody[B]
```
but sometimes we get in response not an informative body, but some answer. So, this Message type is for that.

myazinn: Thank you for such a detailed answer, but it does not contain the answer to my question. 
The presence of field `text` in this class requires that field with such name MUST exist in the response. Are you sure that this field exists in all responses where you use this`Message`?
You correctly pointed out that this is kinda mock for a few cases where you don't care about response body. I think it must be mentioned in the class name, rather than the fact that this is a message of the success response.

OstapStrashevskii: 1) no, field "text" is just my "interpretation". I have no sense about "text" field in real GitLab response
2) Does the name "SuccessResponseMessage" fit, in your opinion?

myazinn: 1) In that case, you should not use a class with such a structure.
2) In my opinion, it doesn't fit. There is no way that we can understand that this is a mock object for the response rather than a real one. For example, names like `DummyResponse`, `MockResponse`, `EmptyResponse`  may give a hint to another developer that something is wrong with such responses and that he should not use it whenever he wants.
If you think that in your cases the response is just a string, that you should do something like that
`post[String, OutputType](???)`
`json` that consists of one string is **not** the same thing as `jsob` object that contains a single string field.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: Excuse me, are you sure about that? Can you give a little explaination?(some tests failed with that change)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: Corrected. Very useful correction, thanks! I will try to use this later! (squash matching where possible)

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
myazinn: What tests have failed? What is the reason of it?
I can write you an explanation, but I suggest you use both of these formatters (before and after changes) and compare the results.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: Have done

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: `Request was rejected with rejection MalformedRequestContentRejection(JsResultException(errors:List((,List(JsonValidationError(List("folder/test.txt" is not an object),WrappedArray())))))`
in **ProjectFileControllerTest:** when **when upload file**
It seems to me that it needs 'path', and jsObject has this structure. But if you sure, I shall try to dig details

myazinn: Look at my first comment.
**_Don't forget to change reads in a similar way_**
Did you do it?

OstapStrashevskii: I guess I know what you talk about. I will check it

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: I think you did not understand me. This class is just a wrap for responseBody. It is not a mock. While I have parametrized Response and parametrized ResponseBody I have a new duty - to pack my success response body into any kind of "wrap". 

```
case class Response[B](status: Int, body: ResponseBody[B], headers: Map[String, Seq[String]])

sealed trait ResponseBody[B]

final case class SuccessResponseBody[B](bodyObject: B) extends ResponseBody[B]
```
The "Message" is just a wrap, maybe the best name for it will be - "SuccessResponseBodyContent" - its too long, but it clear what it does

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: got it

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: Did it

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,65 +4,93 @@ import akka.actor.ActorSystem
...
 import akka.http.scaladsl.model.Uri.Query
 import akka.http.scaladsl.model.headers.RawHeader
-import akka.http.scaladsl.model.{ HttpMethods, HttpRequest, HttpResponse, Uri }
+import akka.http.scaladsl.model.{ HttpEntity, HttpMethods, HttpRequest, HttpResponse, Uri }
 import akka.stream.ActorMaterializer
-import cromwell.pipeline.service.{ HttpClient, Response }
+import cromwell.pipeline.service.{ FailureResponseBody, HttpClient, Response, SuccessResponseBody }
+import play.api.libs.json.{ JsValue, Json, Reads, Writes }
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
```
myazinn, line 12: Imports are not formatted properly (wrong order)

myazinn, line 30: Imports are not formatted properly (wrong order)

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -4,65 +4,93 @@ import akka.actor.ActorSystem
...
+      } else {
+        failureResponse(response)
       }
     }
 
+  private def successResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
+    Unmarshal(response.entity).to[B].map { b =>
+      println("b=" + b)
```
myazinn: What is it doing here?
Please review your code before pushing it.

project/Dependencies.scala
```diff
@@ -56,8 +57,10 @@ object Dependencies {
   lazy val akkaDependencies = Seq(akkaActor, akkaStreams, akkaHttp)
   lazy val jsonDependencies = Seq(playJson, akkaHttpJson, jwtCore)
   lazy val dbDependencies = Seq(slick, slickPg, hikariCP, postgresql, liquibase, yaml)
+//  lazy val testDependencies =
+//    Seq(mockito, akkaTestKit, akkaHttpTestKit, scalaCheck, scalaTestWithIt, scalaMock, logback, wireMock, pegdown)
```
myazinn: Do not keep commented lines in your code unless they have a purpose.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -72,19 +72,27 @@ object Commit {
...
-  implicit object ProjectFileFormat extends Format[ProjectFile] {
-    override def reads(json: JsValue): JsResult[ProjectFile] =
-      JsSuccess(ProjectFile(Paths.get((json \ "path").as[String]), (json \ "content").as[String]))
 
-    override def writes(o: ProjectFile): JsValue = JsObject(
-      Seq("path" -> JsString(o.path.toString), "content" -> JsString(o.content))
-    )
+  implicit object pathFormat extends Format[Path] {
+    override def writes(o: Path): JsValue = JsString(o.toString)
+    override def reads(json: JsValue): JsResult[Path] = JsSuccess(Paths.get(json.as[String]))
```
myazinn: This line is equivalent to something like this.
```
def parseLong(longStr: String): Try[Long] = Success(longStr.toLong)
```
I assume you understand why this is bad.
Use `json.validate` instead of `json.as` and do not wrap it manually into a `JsSuccess`.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -125,3 +127,8 @@ object ProjectUpdateFileRequest {
       unlift(ProjectUpdateFileRequest.unapply)
     )
 }
+
+final case class DummyResponseBody(dummyJsonParameter: String)
```
myazinn: I'm going to ask you the same question as before.
Are you sure that field with a name `dummyJsonParameter` exists in every place where you use this `DummyResponseBody` ?

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import org.scalatest.{ BeforeAndAfterAll, Suite }
+
+trait AkkaTestSources extends BeforeAndAfterAll {
+  this: Suite =>
+//  lazy val actorSystem = ActorSystem("test")
+  implicit lazy val actorSystem = ActorSystem("test")
+//  implicit lazy val materializer: ActorMaterializer = ActorMaterializer()(actorSystem)
```
myazinn: Remove commented lines

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
```diff
@@ -0,0 +1,24 @@
...
+  implicit lazy val materializer: ActorMaterializer = ActorMaterializer()
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+//    actorSystem
```
myazinn: remove it

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -45,40 +61,60 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
...
+        )
+
+        println("gitLabProjectVersioning=" + gitLabProjectVersioning)
+        println("withRepoProject=" + withRepoProject)
+        println(
+          "getProjectVersions(withRepoProject)=" + gitLabProjectVersioning.getProjectVersions(
+            withRepoProject
+          )
+        )
+        println()
```
myazinn: Why are there so many `println`s?

utils/src/main/scala/cromwell/pipeline/utils/json/DummyObject.scala
```diff
@@ -0,0 +1,9 @@
+package cromwell.pipeline.utils.json
```
myazinn: I think we can place this class in `utils` package instead of `utils.json`. There is nothing specific for `json`.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -21,30 +21,39 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
             }
-        case Response(_, body, _) => Future.successful(Left(VersioningException(body)))
+        case Response(_, body, _) =>
+          body match {
+            case SuccessResponseBody(bodyInfo) =>
+              Future.successful(
+                Left(VersioningException("status codes processing may not correctly implemented: " + bodyInfo))
+              )
+            case FailureResponseBody(error) => Future.successful(Left(VersioningException(error)))
+          }
```
myazinn: Remove this nested pattern matching as you did in other places.

myazinn (changes requested):
Please review your code before sending it to review.
It should not contain commented lines or things like `println` unless there is a reason,

project/Dependencies.scala
OstapStrashevskii: I wonder why it is here I double-checked that. So, sorry for that. Done

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
OstapStrashevskii: Looks like I removed all of this, but have forgotten to commit or add this to index!

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: This is just a dummy parameter in dummy object. We can delete GitLab methods implementations in GitLabProjectVersioning and their tests and forget about this, cause this is not fair implementations but just simulations

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: json.validate needs formatter itself, method validate needs implicit parameter - formatter to work

myazinn: So what? Take a look at the implementation of the `.as` method. If you can use `.as` method, then you can also use `.validate` method without a problem.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
myazinn: I don't understand, why we should delete methods in `GitLabProjectVersioning` or tests.

OstapStrashevskii: dummy is just a Dummy, this parameter and object is only used now by GitLabProjectVersioning and GitLabProjectVersioningTest, due to the abstractness of these class methods

utils/src/test/scala/cromwell/pipeline/utils/AkkaTestSources.scala
OstapStrashevskii: removed

services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
OstapStrashevskii: deleted, I will be carefull after resolving conflicts =(

utils/src/main/scala/cromwell/pipeline/utils/json/DummyObject.scala
OstapStrashevskii: done, moved

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: done

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: agreed, ofcourse, done

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: deleted

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: done

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
myazinn: You are right, but I'm struggling to understand how does it related to my original question.
```suggestion
final case class DummyResponseBody()
```
That's what I wanted from you. An empty class without fields. In that case, we won't fail if the response from real GitLab services does not contain `test` or `dummyJsonParameter` fields. As far as I remember, I told you to make such an empty class for it when we discussed it in Google Meetings.

OstapStrashevskii: the only reason why this dummuobject are not in test utils package is that it used in simulation-methods post and put (GitLabProjectVersioning class)

OstapStrashevskii: Done!

myazinn (approved):
Well done!

GrigoriiBerezin (approved):
Looks good to me. Thanks for your masterpiece!

iliaonishchenko (approved):
Nice work!
Please squash the commits.

build.sbt
```diff
@@ -86,7 +86,7 @@ lazy val utils =
   (project in file("utils"))
     .configs(IntegrationTest)
     .settings(
-      libraryDependencies ++= (jsonDependencies ++ testContainers ++ coreTestDependencies ++ dbDependencies) :+ configHokon :+ cats :+ playFunctional :+ pegdown,
+      libraryDependencies ++= (jsonDependencies ++ testContainers ++ coreTestDependencies ++ dbDependencies) :+ scalaTest :+ configHokon :+ cats :+ playFunctional :+ pegdown,
```
myazinn: `scalaTest` already included in `coreTestDependencies`

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
```diff
@@ -1,69 +1,138 @@
...
+  /**
+   * Parse and validate response body
+   *
+   * @return SuccessResponseBody(value) if body json is valid or FailureResponseBody otherwise
+   */
+  private def validateResponse[B](
+    response: HttpResponse
+  )(implicit ec: ExecutionContext, f: Reads[B]): Future[Response[B]] =
+    response.entity.toStrict(expirationTime).map { body =>
+      Json.parse(body.data.utf8String).validate[B] match {
```
myazinn: I assume there is a reason to use `Json.pase` explicitly instead of `Unmarshal(entity).to[B]`.
I'm just wondering what the reason is.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -66,7 +66,18 @@ object ProjectUpdateRequest {
...
+
+    override def writes(o: GitLabVersion): JsValue = JsString(o.toString)
+
+    override def reads(json: JsValue): JsResult[GitLabVersion] = for {
+      name <- (json \ "name").validate[PipelineVersion]
+      message <- (json \ "message").validate[String]
+      target <- (json \ "target").validate[String]
+      commit <- (json \ "commit").validate[Commit]
+    } yield GitLabVersion(name, message, target, commit)
+  }
```
myazinn: Why did you change this formatter?
Also, have you tried to serialize and then deserialize an instance of `GitLabVersion` with this formatter? It looks like it won't work.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -188,4 +198,36 @@ object ProjectUpdateFileRequest {
...
+  implicit lazy val successResponseMessageFormat: OFormat[SuccessResponseMessage] =
+    Json.format[SuccessResponseMessage]
+}
+
+/**
+ * Wrapper type contains empty response message to use in [[cromwell.pipeline.service.HttpClient]]
+ *
+ * @param message empty payload message
+ */
+final case class EmptyPayload private (message: String = "")
```
myazinn: It makes no sense to create a `case class` with a private constructor. Anyone can create an instance using `EmptyPayload.apply` method.

services/src/main/scala/cromwell/pipeline/service/HttpClient.scala
```diff
@@ -1,17 +1,74 @@
...
+    bf: Reads[B],
+    pf: Writes[P]
+  ): Future[Response[B]]
 }
 
-case class Response(status: Int, body: String, headers: Map[String, Seq[String]])
+/**
+ * Contains response data with unmarshalled Json
+ *
+ * @param status equals Akka StatusCode
```
myazinn: I think you should remove any mention of Akka here. It is not a part of the HTTP client contract. It is an implementation detail that may change.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -175,13 +170,16 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
-            tagFile <- tagsFiles
-            if tagFile.commitId == tagProject.commit.id
-          } yield tagProject)
-        case (_, Left(exception)) => Left(VersioningException(exception.message))
-        case (Left(exception), _) => Left(VersioningException(exception.message))
+          Right(
+            tagsProject.flatMap(
+              tagProject =>
+                tagsFiles.withFilter(tagFile => tagFile.commitId == tagProject.commit.id).map(tagFile => tagProject)
+            )
```
myazinn: Are you sure that it's better than it was? :)
Maybe it's just me, but this code is much harder to read compared to its version on the right. @ilyaonishenko what do you think?

myazinn (changes requested):
You've added documentation, this is cool!
Unfortunately, you've also added new problems :( Please, take a look at my comments and fix them.

controllers/src/main/scala/cromwell/pipeline/controller/AkkaHttpClient.scala
OstapStrashevskii: The only reason forced me to change this was: separating parse and validate logic, and the possibility of throwing different, particular exceptions in the cases where json can be parsed but not validated. I know, that akka's marshal/umarshall uses playjson writes and reads. And I know that validate here is just "reads"(if observe playjson sources). But I made this only to extract reads to get a particular exception. Maybe it is not necessary to do it, but I am just refactoring code while resolving newer and newer conflicts, and everybody has this particular validation check in his request, and I try to be polite & careful and to save more as I can save.)) Maybe I have to strengthen my refactoring politics? ))

myazinn: Maybe `Future#transform` is the method that will solve your problem?
`Unmarshal(entity).to[B].transform { }` will allow you to separate different exceptions and recover with `FailureResponseBody` if necessary.

OstapStrashevskii: Thank you,  excellent idea! I think this is just what I need.

build.sbt
OstapStrashevskii: done

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: Changed back. Unfortunately, I forgot why. There is no proper reason to change it.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
OstapStrashevskii: done, modifier removed

services/src/main/scala/cromwell/pipeline/service/HttpClient.scala
OstapStrashevskii: akka mentions erased from trait

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
OstapStrashevskii: I like maps. But for sure for-c is better here. Done.

myazinn (approved):
You made this PR great again!

PR #79, opened 2020-06-09
EPMLSTRCMW-161_CREATE_DTO_VERSION_FOR_PIPELINE

GrigoriiBerezin:
Divide Version onto PipelineVersion and GitLabVersion.
GitLabVersion contains full information of tag and PipelineVersion contains only name of tag.
PipelineVersion was added to friendly input for users, by PipelineVersion we can get GitLabVersion

### DIFF
--- a/model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
+++ b/model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
@@ -1,0 +1,33 @@
+package cromwell.pipeline.model.wrapper
+
+import cats.data.{ NonEmptyChain, Validated }
+import cats.implicits.catsStdShowForString
+import cromwell.pipeline.model.validator.Wrapped
+import play.api.libs.json.Format
+
+final class VersionValue private (override val unwrap: Int) extends AnyVal with Wrapped[Int]
+
+object VersionValue extends Wrapped.Companion {
+  type Type = Int
+  type Wrapper = VersionValue
+  type Error = String
+  val pattern = "^[0-9]+$"
+  implicit val versionNumberFormat: Format[VersionValue] = wrapperFormat
+  def increment(value: VersionValue): VersionValue = create(value.unwrap + 1)
+  def fromString(value: String): ValidationResult[Wrapper] =
+    validateString(value) match {
+      case Validated.Valid(content)  => Validated.Valid(create(content.toInt))
+      case Validated.Invalid(errors) => Validated.Invalid(errors)
+    }
+  protected def validateString(value: String): ValidationResult[String] = Validated.cond(
+    value.matches(pattern),
+    value,
+    NonEmptyChain.one("Value should be not negative number")
+  )
+  override protected def create(value: Int): VersionValue = new VersionValue(value)
+  override protected def validate(value: Int): ValidationResult[Int] = Validated.cond(
+    value >= 0,
+    value,
+    NonEmptyChain.one("Value should be not negative")
+  )
+}
--- a/portal/src/main/resources/application.conf
+++ b/portal/src/main/resources/application.conf
@@ -44,5 +44,7 @@ database {
     token = ${?CMWL_GITLAB_TOKEN}
     defaultFileVersion = "v0.0.1"
     defaultFileVersion = ${?CMWL_DEFAULT_FILE_VERSION}
+    defaultBranch = "master"
+    defaultBranch = ${?CMWL_DEFAULT_BRANCH}
   }
 }
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
@@ -2,9 +2,11 @@ package cromwell.pipeline.datastorage.dto
 
 import java.nio.file.{ Path, Paths }
 
+import cats.data.Validated
+import cats.implicits._
+import cromwell.pipeline.model.wrapper.{ UserId, VersionValue }
 import play.api.libs.functional.syntax._
 import play.api.libs.json._
-import cromwell.pipeline.model.wrapper.UserId
 import slick.lifted.MappedTo
 
 final case class Project(
@@ -62,9 +64,52 @@ object ProjectUpdateRequest {
   implicit val updateRequestFormat: OFormat[ProjectUpdateRequest] = Json.format[ProjectUpdateRequest]
 }
 
-final case class Version(name: String, message: String, target: String, commit: Commit)
-object Version {
-  implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
+final case class GitLabVersion(name: PipelineVersion, message: String, target: String, commit: Commit)
+object GitLabVersion {
+  implicit val gitlabVersionFormat: OFormat[GitLabVersion] = Json.format[GitLabVersion]
+}
+
+final case class PipelineVersion(major: VersionValue, minor: VersionValue, revision: VersionValue)
+    extends Ordered[PipelineVersion] {
+  import VersionValue._
+  def name: String = s"v$major.$minor.$revision"
+
+  private val ordering: Ordering[PipelineVersion] = Ordering.by(v => (v.major, v.minor, v.revision))
+
+  override def compare(that: PipelineVersion): Int =
+    ordering.compare(this, that)
+
+  def increaseMajor: PipelineVersion =
+    this.copy(major = increment(this.major))
+
+  def increaseMinor: PipelineVersion =
+    this.copy(minor = increment(this.minor))
+
+  def increaseRevision: PipelineVersion =
+    this.copy(revision = increment(this.revision))
+
+  override def toString: String = this.name
+}
+
+object PipelineVersion {
+  import VersionValue._
+
+  private val pattern = "^v(.+)\\.(.+)\\.(.+)$".r
+  def apply(versionLine: String): PipelineVersion =
+    versionLine match {
+      case pattern(v1, v2, v3) =>
+        val validationResult = (fromString(v1), fromString(v2), fromString(v3)).mapN(PipelineVersion.apply)
+        validationResult match {
+          case Validated.Valid(content)  => content
+          case Validated.Invalid(errors) => throw PipelineVersionException(errors.toString)
+        }
+      case _ => throw PipelineVersionException(s"Format of version name: 'v(int).(int).(int)', but got: $versionLine")
+    }
+
+  final case class PipelineVersionException(message: String) extends Exception
+
+  implicit val pipelineVersionFormat: Format[PipelineVersion] =
+    implicitly[Format[String]].inmap(PipelineVersion.apply, _.name)
 }
 
 final case class Commit(id: String)
@@ -115,12 +160,12 @@ object FileContent {
   implicit lazy val validateFileRequestFormat: OFormat[FileContent] = Json.format[FileContent]
 }
 
-final case class ProjectUpdateFileRequest(project: Project, projectFile: ProjectFile, version: Option[Version])
+final case class ProjectUpdateFileRequest(project: Project, projectFile: ProjectFile, version: Option[PipelineVersion])
 
 object ProjectUpdateFileRequest {
   implicit lazy val projectUpdateFileRequestFormat: OFormat[ProjectUpdateFileRequest] =
     ((JsPath \ "project").format[Project] ~ (JsPath \ "projectFile").format[ProjectFile] ~ (JsPath \ "version")
-      .formatNullable[Version])(
+      .formatNullable[PipelineVersion])(
       ProjectUpdateFileRequest.apply,
       unlift(ProjectUpdateFileRequest.unapply)
     )
--- a/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
+++ b/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
@@ -5,8 +5,11 @@ import java.util.UUID
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
 
+import scala.util.Random
+
 object TestProjectUtils {
 
+  private def randomInt(range: Int): Int = Random.nextInt(range)
   private def randomUuidStr: String = UUID.randomUUID().toString
   def getDummyProjectId: ProjectId = ProjectId(randomUuidStr)
   def getDummyRepository: Repository = Repository(s"repo-$randomUuidStr")
@@ -19,10 +22,16 @@ object TestProjectUtils {
     visibility: Visibility = Private
   ): Project = Project(projectId, ownerId, name, active, repository, visibility)
   def getDummyCommit(id: String = randomUuidStr): Commit = Commit(id)
-  def getDummyVersion(
-    name: String = s"name-$randomUuidStr",
+  def getDummyPipeLineVersion(
+    v1: Int = 1 + randomInt(12),
+    v2: Int = 1 + randomInt(12),
+    v3: Int = 1 + randomInt(12)
+  ): PipelineVersion =
+    PipelineVersion(s"v$v1.$v2.$v3")
+  def getDummyGitLabVersion(
+    version: PipelineVersion = getDummyPipeLineVersion(),
     message: String = s"message-$randomUuidStr",
     target: String = s"target-$randomUuidStr",
     commit: Commit = getDummyCommit()
-  ): Version = Version(name, message, target, commit)
+  ): GitLabVersion = GitLabVersion(version, message, target, commit)
 }
--- a/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
@@ -4,7 +4,7 @@ import java.net.URLEncoder
 import java.nio.file.Path
 
 import cromwell.pipeline.datastorage.dto.File.UpdateFileRequest
-import cromwell.pipeline.datastorage.dto.{ Project, ProjectFile, Version }
+import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.utils.{ GitLabConfig, HttpStatusCodes }
 import play.api.libs.json.{ JsError, JsResult, JsSuccess, Json }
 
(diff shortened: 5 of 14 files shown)


services/src/test/scala/cromwell/pipeline/service/GitLabProjectVersioningTest.scala
```diff
@@ -151,7 +151,7 @@ class GitLabProjectVersioningTest extends AsyncWordSpec with ScalaFutures with M
       "return file with 200 response" taggedAs Service in {
         val project = TestProjectUtils.getDummyProject()
         val path = Paths.get("test.md")
-        val version = dummyVersion;
+        val version = PipelineVersion("version-name")
```
OstapStrashevskii, line 154: Is "version-name" a proper variant of string you use in PipelineVersion? Maybe it would be better to use version-like strings as in all other tests? This thing is quite misleading.

portal/src/main/resources/application.conf
```diff
@@ -42,7 +42,9 @@ database {
     path = ${?CMWL_GITLAB_USER_NAME}
     token = "token"
     token = ${?CMWL_GITLAB_TOKEN}
-    defaultFileVersion = "v0.0.1"
+    defaultFileVersion = "v.0.0.1"
     defaultFileVersion = ${?CMWL_DEFAULT_FILE_VERSION}
+    defaultBranch = "master"
+    defaultBranch = ${?CMWL_DAFAUTL_BRANCH}
```
IaroslavTavchenkov, line 48: Hey!  You got a typo.
You wrote DAFAUTL instead of DEFAULT

GrigoriiBerezin: Thx, will fix it asap

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-    val fileUrl = s"${config.url}projects/${project.repository}/repository/files/$path"
-    val versionValue = version.map(_.name).getOrElse(config.defaultFileVersion)
+    val repositoryId: Repository = project.repository match {
+      case Some(repository) => repository
+      case None             => throw VersioningException(s"No repository for project: $project")
+    }
+    val fileUrl = s"${config.url}projects/${repositoryId.value}/repository/files/$path"
 
+    getProjectVersions(project).flatMap(versions => {
+      val latestVersion: Either[VersioningException, PipelineVersion] = (version, versions) match {
```
ilyaonishenko, line 27: Why we're taking into account new version, coming from user, if the variable is `latestVersion`?

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      }
+      latestVersion match {
+        case Left(error) =>
+          Future.successful(Left(error))
+        case Right(tagName) =>
+          val payload =
+            Json.stringify(Json.toJson(UpdateFileRequest(projectFile.content, tagName.toString, config.defaultBranch)))
+          for {
+            updateResult <- httpClient.put(fileUrl, payload = payload, headers = config.token)
+            createResult <- httpClient.post(fileUrl, payload = payload, headers = config.token)
```
ilyaonishenko, line 45: What?

ilyaonishenko: Why?

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+            tagResult <- createTag(repositoryId, tagName)
+          } yield (updateResult, createResult, tagResult) match {
+            case (Response(HttpStatusCodes.OK, _, _), Response(HttpStatusCodes.BadRequest, _, _), Right(_)) =>
```
ilyaonishenko, line 48: You're suppressing message from GitLab. But it might be useful. Might be more useful than `Success update file`

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-        fileUrl,
-        payload = Json.stringify(
-          Json.toJson(UpdateFileRequest(versionValue, projectFile.content, versionValue))
-        ),
-        headers = config.token
+      .post(
+        tagUrl,
+        params = Map("tag_name" -> version.name, "ref" -> config.defaultBranch),
+        headers = config.token,
+        payload = ""
```
ilyaonishenko, line 68: Please create a variable `EmptyPayload`

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -62,9 +63,50 @@ object ProjectUpdateRequest {
...
 
-final case class Version(name: String, message: String, target: String, commit: Commit)
-object Version {
-  implicit val versionPlayFormat: OFormat[Version] = Json.format[Version]
+final case class GitLabVersion(name: PipelineVersion, message: String, target: String, commit: Commit)
+object GitLabVersion {
+  implicit val gitlabVersionFormat: OFormat[GitLabVersion] = Json.format[GitLabVersion]
+}
+
+final case class PipelineVersion private (versions: Int*) extends Ordered[PipelineVersion] {
```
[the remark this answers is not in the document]
GrigoriiBerezin: Ok

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -62,9 +63,50 @@ object ProjectUpdateRequest {
...
+    comparableV.find(i => i != 0) match {
+      case Some(value) => value
+      case None        => this.versions.length.compare(that.versions.length)
+    }
+  }
+
+  def changeVersion(index: Int, value: Int): PipelineVersion =
+    if (index >= this.versions.length) throw PipelineVersionException("Out of bounds")
+    else if (value <= 0) throw PipelineVersionException("Version must be positive value")
+    else new PipelineVersion(this.versions.updated(index, value): _*)
```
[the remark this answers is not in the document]
GrigoriiBerezin: Ok, I will delete it, because there will be a lot of methods like increase something in case that we have a seq of versions under the hood.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+    }
+    val fileUrl = s"${config.url}projects/${repositoryId.value}/repository/files/$path"
 
+    getProjectVersions(project).flatMap(versions => {
+      val latestVersion: Either[VersioningException, PipelineVersion] = (version, versions) match {
+        case (Some(_), Left(error))                             => Left(error)
+        case (Some(tagName), Right(v :: _)) if v.name < tagName => Right(tagName)
+        case (Some(tagName), Right(v :: _)) if v.name >= tagName =>
+          Left(VersioningException(s"Your version $tagName is out of date. Current version of project: ${v.name}"))
+        case (Some(_), Right(_))   => Right(PipelineVersion(config.defaultFileVersion))
```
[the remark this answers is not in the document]
GrigoriiBerezin: Because it can be a situation where there is no versions yet in project, than we should take our version.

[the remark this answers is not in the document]
ilyaonishenko: But you're taking the default version here.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Didn't catch your question - latestVersion here is really latest version between project versions and user version.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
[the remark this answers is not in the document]
ilyaonishenko: A lot of methods is not an issue.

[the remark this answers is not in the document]
ilyaonishenko: As far as they continue to be useful.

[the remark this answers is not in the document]
GrigoriiBerezin: It probably will be infinite methods because the size of Seq of Versions is unlimited :)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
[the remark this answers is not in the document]
GrigoriiBerezin: Yep, thx, fix it already :)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: I thought that it will look more accurate than chain of try to updateFile and if this file doesn't exist, createFile, but will fix it, no problem

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Will do it, thx

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Yes, that's true :)

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,63 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      case None             => throw VersioningException(s"No repository for project: $project")
+    }
+    val fileUrl = s"${config.url}projects/${repositoryId.value}/repository/files/$path"
 
+    getProjectVersions(project).flatMap(versions => {
+      val latestVersion: Either[VersioningException, PipelineVersion] = (version, versions) match {
+        case (Some(_), Left(error))                             => Left(error)
+        case (Some(tagName), Right(v :: _)) if v.name < tagName => Right(tagName)
+        case (Some(tagName), Right(v :: _)) if v.name >= tagName =>
+          Left(VersioningException(s"Your version $tagName is out of date. Current version of project: ${v.name}"))
```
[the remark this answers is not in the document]
GrigoriiBerezin: Fixed it, thx

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,73 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+  override def updateFile(project: Project, projectFile: ProjectFile, version: Option[PipelineVersion])(
     implicit ec: ExecutionContext
   ): AsyncResult[String] = {
     val path = URLEncoder.encode(projectFile.path.toString, "UTF-8")
-    val fileUrl = s"${config.url}projects/${project.repository}/repository/files/$path"
-    val versionValue = version.map(_.name).getOrElse(config.defaultFileVersion)
+    val repositoryId: Repository = project.repository match {
+      case Some(repository) => repository
+      case None             => throw VersioningException(s"No repository for project: $project")
+    }
```
myazinn, line 23:
```suggestion
    val repositoryId: Repository = project.repository.getOrElse(throw VersioningException(s"No repository for project: $project")
```
`getOrElse` is better here since it explicitly says what are you want to do.
Also, maybe you should return `Future.failed` instead of throwing an exception?

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,73 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+    val fileUrl = s"${config.url}projects/${repositoryId.value}/repository/files/$path"
+
+    getProjectVersions(project).flatMap(versions => {
+      val latestVersion: Either[VersioningException, PipelineVersion] = (version, versions) match {
+        case (Some(_), Left(error)) => Left(error)
+        case (Some(userVersion), Seq(headVersion: GitLabVersion, _)) if headVersion.name >= userVersion =>
```
myazinn, line 29: If I'm not mistaken, you are comparing here strings instead of versions. I.e. you are not using `compare` method that you implemented.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,73 @@ import scala.concurrent.{ ExecutionContext, Future }
...
     val path = URLEncoder.encode(projectFile.path.toString, "UTF-8")
-    val fileUrl = s"${config.url}projects/${project.repository}/repository/files/$path"
-    val versionValue = version.map(_.name).getOrElse(config.defaultFileVersion)
+    val repositoryId: Repository = project.repository match {
+      case Some(repository) => repository
+      case None             => throw VersioningException(s"No repository for project: $project")
+    }
+    val fileUrl = s"${config.url}projects/${repositoryId.value}/repository/files/$path"
+
+    getProjectVersions(project).flatMap(versions => {
```
myazinn, line 26:
```suggestion
    getProjectVersions(project).flatMap { versions => 
```

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
myazinn: I'm kinda surprised that it compiled, but it looks like the compiler did try to stop you. That's why you forced him to think that `headVersion` is a `GitLabVersion` instead of `Any`.
`versions` is an `Either`. If you want to pattern match on it, you must write `Right(Seq(...))`.
Same for line 36 below.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,73 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+                case Response(HttpStatusCodes.OK, body, _) =>
+                  createTag(repositoryId, newVersion).map {
+                    case Right(_)        => Right(body)
+                    case Left(exception) => Left(exception)
+                  }
+                case Response(_, body, _) => Future.successful(Left(VersioningException(body)))
+              }
+          }
+      }
+    })
```
myazinn, line 65: There are so many things that happen inside a single method. Maybe you should separate it into smaller methods like `validateVersion`, `sendUpdateVersionRequest`, or something like that.

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -5,8 +5,11 @@ import java.util.UUID
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
 
+import scala.util.Random
+
 object TestProjectUtils {
 
+  private def randomInt: Int = Random.nextInt()
```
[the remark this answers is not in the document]
GrigoriiBerezin: Fixed it

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Ok, thx

GrigoriiBerezin: About not using compare method: `headVersion` is `GitLabVersion` type, that has param `name`  of `PipelineVersion` type

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Thx

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -62,9 +62,47 @@ object ProjectUpdateRequest {
...
+  implicit val gitlabVersionFormat: OFormat[GitLabVersion] = Json.format[GitLabVersion]
+}
+
+final case class PipelineVersion(major: VersionValue, minor: VersionValue, revision: VersionValue)
+    extends Ordered[PipelineVersion] {
+  import VersionValue._
+  def name: String = s"v$major.$minor.$revision"
+
+  private val ordering: Ordering[PipelineVersion] =
+    Ordering[(VersionValue, VersionValue, VersionValue)].on((pv: PipelineVersion) => (pv.major, pv.minor, pv.revision))
```
myazinn, line 76:
```suggestion
    Ordering[PipelineVersion].by
```

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -62,9 +62,47 @@ object ProjectUpdateRequest {
...
+
+  def increaseRevision: PipelineVersion =
+    this.copy(revision = increment(this.revision))
+
+  override def toString: String = this.name
+}
+
+object PipelineVersion {
+  import VersionValue._
+  private val pattern = """^v(\d+)\.(\d+)\.(\d+)$""".r
```
myazinn, line 95: This pattern assumes the inner structure of the `VersionValue`. However, this structure must be hidden even for `PipelineVersion`
You must check that the string starts with `v` and contains three subgroups separated by `.`. Each of these subgroups must be validated inside `VersionValue.fromString` method, rather than here.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,88 @@ import scala.concurrent.{ ExecutionContext, Future }
 class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     extends ProjectVersioning[VersioningException] {
 
-  override def updateFile(project: Project, projectFile: ProjectFile, version: Option[Version])(
+  private def getLastProjectVersion(project: Project)(implicit ec: ExecutionContext): Future[Option[PipelineVersion]] =
+    getProjectVersions(project).map {
+      case Left(_)                          => None
```
myazinn, line 18: `Left` means we got some errors. I don't think we should swallow that error and pretend that project doesn't have any versions.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,88 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      case Right(Seq(v: GitLabVersion, _*)) => Some(v.name)
+      case Right(Seq())                     => Some(PipelineVersion(config.defaultFileVersion))
+    }
+
+  private def compareVersions(
+    optionProjectVersion: Option[PipelineVersion],
+    optionUserVersion: Option[PipelineVersion]
+  )(implicit ec: ExecutionContext): Either[VersioningException, PipelineVersion] =
```
myazinn, line 26: I can't find the place where you need `ExecutionContext` in this method.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,88 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+    (optionProjectVersion, optionUserVersion) match {
+      case (Some(projectVersion), Some(userVersion)) =>
+        if (projectVersion >= userVersion)
+          Left(
+            VersioningException(
+              s"Your version $userVersion is out of date. Current version of project: $projectVersion"
+            )
+          )
+        else Right(userVersion)
+      case (Some(projectVersion), None) => Right(projectVersion.increaseRevision)
```
myazinn, line 36: Why does method `compareVersions` returns increased version? What is the purpose of this method?
Either change method behavior or method name.

model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
```diff
@@ -0,0 +1,50 @@
...
+      case Validated.Valid((v1, v2, v3)) => Validated.Valid((create(v1.toInt), create(v2.toInt), create(v3.toInt)))
+      case Validated.Invalid(errors)     => Validated.Invalid(errors)
+    }
+  protected def validateFullVersion(value: String): ValidationResult[StringGroup] = Validated.cond(
+    value.matches(fullVersionPattern.regex),
+    value match {
+      case fullVersionPattern(v1, v2, v3) => (v1, v2, v3)
+    },
+    NonEmptyChain.one(s"Format of version name: 'v(int).(int).(int)', but got: $value")
+  )
```
myazinn, line 33: What are you doing?
I told that `VersionValue` is a **single** version. You must not parse a full version here, this the responsibility of the `PipelineVersion` method!

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,89 @@ import scala.concurrent.{ ExecutionContext, Future }
 class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     extends ProjectVersioning[VersioningException] {
 
-  override def updateFile(project: Project, projectFile: ProjectFile, version: Option[Version])(
+  private def getLastProjectVersion(project: Project)(implicit ec: ExecutionContext): Future[Option[PipelineVersion]] =
+    getProjectVersions(project).map {
+      case Left(error)                      => throw error
```
myazinn, line 18: Throwing an exception is not the Scala way, you should avoid it wherever possible.
It is better to return `Future.failed` instead.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -13,38 +13,89 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+              case Response(_, body, _) => Future.successful(Left(VersioningException(body)))
+            }
+        }
+    }
+  }
+
+  private def createTag(projectId: Repository, version: PipelineVersion)(
+    implicit ec: ExecutionContext
+  ): AsyncResult[String] = {
+    val emptyPayload = "";
```
myazinn, line 87:
```suggestion
    val emptyPayload = ""
```
Didn't notice it last time.

GrigoriiBerezin: Omg, strange mistake from me

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
GrigoriiBerezin: Yep, will change it

model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
GrigoriiBerezin: As I get it from a previous message, this should be in micro type. In my previous commit validate form of full version was in PipelineVersion

myazinn: Microtype `VersionValue` must contain information about parsing a single value into a `VersionValue` and nothing more.
`PipelineVersion` must know that it consists of three `VersionValue`s.
In your previous commit, `PipelineVersion` also knew that `VersionValue` consists of digits, which is wrong.
Now `VersionValue` knows that `PipelineVersion` consists of three `VersionValues`, which is also wrong.

myazinn: `PipelineVersion` must take a full string, validate that it matches the following regex (or something similar, idk)
`^v([^.]*).([^.]*).([^.]*)$`
If it matches then it passes each of these three elements into `VersionValue.fromString` method. That's all.

PR #80, opened 2020-06-22
EPMLSTRCMW-164-Implement-getFileTree-method

manvicha:
feat: Implement getFiles Tree method

The method scans the project structure for the
specified version (if version is not specified,
then uses the latest one) and returns a list 
which describes location of all files stored
in the project

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
@@ -190,9 +190,35 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     implicit ec: ExecutionContext
   ): AsyncResult[List[GitLabVersion]] = ???
 
-  override def getFileTree(project: Project, version: Option[PipelineVersion])(
+  override def getFilesTree(project: Project, version: Option[PipelineVersion])(
     implicit ec: ExecutionContext
-  ): AsyncResult[List[String]] = ???
+  ): AsyncResult[Seq[FileTree]] = {
+
+    val versionId: Map[String, String] = version match {
+      case Some(version) => Map("ref" -> version.name, "recursive" -> "true")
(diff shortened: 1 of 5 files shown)


myazinn:
This PR contains too many changes :) 
You must change the target branch from master to develop.

myazinn (approved):
Looks good to me. Thank you for your work.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -148,9 +148,33 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
 
-  override def getFileTree(project: Project, version: Option[PipelineVersion])(
+  override def getFilesTree(project: Project, version: Option[PipelineVersion])(
     implicit ec: ExecutionContext
-  ): AsyncResult[List[String]] = ???
+  ): AsyncResult[Seq[FileTree]] = {
+
+    val versionId: String = version match {
+      case Some(version) => version.name
+      case None          => config.defaultFileVersion
```
iliaonishchenko: I guess it's not how it should work.
`defaultFileVersion` actually means `initialFileVersion`... (could you please rename it?)
So it's a version which we're using **if there're no versions in git.** 
In this case we're trying to get version from git, that means that **version already exists**. 

The correct behaviour here if version is `None` is to make a request to git and get the latest version. And if there is no version in git we should return error.

iliaonishchenko (changes requested):

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
manvicha: I did the changes according to your comments. Could you skim and see if i omitted something, please?

iliaonishchenko (approved):
Looks good to me!

PR #83, opened 2020-07-13
feat: Introduce ADT for exceptions in ProjectVersioning

manvicha:
Creating the first draft trying to encapsulate
the exceptions throw by the ProjectVersioning
trait with more details

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
@@ -41,7 +41,10 @@ class ProjectFileController(wdlService: ProjectFileService)(
                   case (Left(_), Right(response))  => StatusCodes.Created.intValue -> response
                   case (_, Left(response))         => response match {
                     case HttpException(message) => StatusCodes.UnprocessableEntity.intValue -> message
-                    case _ => StatusCodes.UnprocessableEntity.intValue -> "Bad request"
+                    case FileException(message) => StatusCodes.UnprocessableEntity.intValue -> message
+                    case GitException(message) => StatusCodes.UnprocessableEntity.intValue -> message
+                    case RepositoryException(message) => StatusCodes.UnprocessableEntity.intValue -> message
+                    case ProjectException(message) => StatusCodes.UnprocessableEntity.intValue -> message
                   }
                 }
               }) {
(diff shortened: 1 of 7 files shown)


controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
```diff
@@ -38,7 +39,10 @@ class ProjectFileController(wdlService: ProjectFileService)(
                 (validateResponse, uploadResponse) match {
                   case (Right(_), Right(response)) => StatusCodes.OK.intValue -> response
                   case (Left(_), Right(response))  => StatusCodes.Created.intValue -> response
-                  case (_, Left(response))         => StatusCodes.UnprocessableEntity.intValue -> response.message
+                  case (_, Left(response))         => response match {
+                    case HttpException(message) => StatusCodes.UnprocessableEntity.intValue -> message
```
manvicha, line 43: Not sure if this has to be handled like this

mnesvit: I believe here we can use exception types in patterns to avoid inner pattern matching:
```case (_, Left(HttpException(message))) =>  StatusCodes.UnprocessableEntity.intValue -> message ```
```case (_, Left(response)) => StatusCodes.UnprocessableEntity.intValue -> "Bad request"```

File: services/src/main/scala/cromwell/pipeline/service/VersioningException.scala [hunk 2]
```diff
@@ -0,0 +1,11 @@
+package cromwell.pipeline.service
+
+sealed abstract class VersioningException extends Exception
```
mnesvit: Maybe it could be useful to make ```message``` a field of ```VersioningException```?

manvicha: It is done

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
manvicha: Thanks!!! I was not sure about include all the exceptions, it makes sense to handle the exception as you proposed

iliaonishchenko (approved):
Looks good to me.

controllers/src/main/scala/cromwell/pipeline/controller/ProjectFileController.scala
myazinn: Why not? They are handled in the same way.
If we decide to handle some of them differently, we can split them again.

mnesvit (approved):

PR #84, opened 2020-07-15
feat: delete endpoint for project configuration

Dmitrii-Shevchenko:
EPMLSTRCMW-166
Implement DELETE endpoint for ProjectConfguration

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectConfigurationController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectConfigurationController.scala
@@ -4,7 +4,8 @@ import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
 import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
-import cromwell.pipeline.datastorage.dto.{ ProjectConfiguration, ProjectId }
+import cromwell.pipeline.datastorage.dto.project.configuration.ProjectConfigurationEntity
+import cromwell.pipeline.datastorage.dto.ProjectId
 import cromwell.pipeline.service.ProjectConfigurationService
 import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
 
@@ -18,7 +19,7 @@ class ProjectConfigurationController(projectConfigurationService: ProjectConfigu
     path("configurations") {
       concat(
         put {
-          entity(as[ProjectConfiguration]) { request =>
+          entity(as[ProjectConfigurationEntity]) { request =>
             onComplete(projectConfigurationService.addConfiguration(request)) {
               case Failure(e)            => complete(StatusCodes.InternalServerError, e.getMessage)
               case Success(updateResult) => complete(updateResult)
@@ -26,16 +27,22 @@ class ProjectConfigurationController(projectConfigurationService: ProjectConfigu
           }
         },
         get {
-          concat(
-            parameter('project_id.as[String]) { projectId =>
-              onComplete(projectConfigurationService.getById(ProjectId(projectId))) {
-                case Failure(e)                   => complete(StatusCodes.InternalServerError, e.getMessage)
-                case Success(Some(configuration)) => complete(configuration)
-                case Success(None) =>
-                  complete(StatusCodes.NotFound, s"There is no configuration with project_id: $projectId")
-              }
+          parameter("project_id".as[String]) { projectId =>
+            onComplete(projectConfigurationService.getById(ProjectId(projectId))) {
+              case Failure(e)                   => complete(StatusCodes.InternalServerError, e.getMessage)
+              case Success(Some(configuration)) => complete(configuration)
+              case Success(None) =>
+                complete(StatusCodes.NotFound, s"There is no configuration with project_id: $projectId")
             }
-          )
+          }
+        },
+        delete {
+          parameter("project_id".as[String]) { projectId =>
+            onComplete(projectConfigurationService.deactivateConfiguration(ProjectId(projectId))) {
+              case Failure(e)            => complete(StatusCodes.InternalServerError, e.getMessage)
+              case Success(updateResult) => complete(updateResult)
+            }
+          }
         }
       )
     }
--- a/controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
@@ -7,6 +7,7 @@ import akka.http.scaladsl.testkit.ScalatestRouteTest
 import cromwell.pipeline.datastorage.dao.repository.utils.{ TestProjectUtils, TestUserUtils }
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.datastorage.dto.project.configuration.ProjectConfigurationEntity
 import cromwell.pipeline.service.ProjectConfigurationService
 import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
 import org.mockito.Mockito.when
@@ -27,12 +28,16 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
         ProjectFileConfiguration(Paths.get("/home/file"), List(FileParameter("nodeName", StringTyped(Some("hello")))))
       )
     )
+    val configurationRequest =
+      ProjectConfigurationEntity(configuration.projectId, configuration.projectFileConfigurations)
+    val configurationResponse =
+      ProjectConfigurationEntity(configuration.projectId, configuration.projectFileConfigurations)
+    val error = new RuntimeException("Something went wrong")
 
     "update configuration" should {
-      val error = new RuntimeException("Something went wrong")
 
       "return success for update configuration" in {
-        when(configurationService.addConfiguration(configuration)).thenReturn(Future.successful("Success"))
+        when(configurationService.addConfiguration(configurationRequest)).thenReturn(Future.successful("Success"))
         Put("/configurations", configuration) ~> configurationController.route(accessToken) ~> check {
           status shouldBe StatusCodes.OK
           entityAs[String] shouldBe "Success"
@@ -40,7 +45,7 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
       }
 
       "return InternalServerError when failure update configuration" in {
-        when(configurationService.addConfiguration(configuration)).thenReturn(Future.failed(error))
+        when(configurationService.addConfiguration(configurationRequest)).thenReturn(Future.failed(error))
         Put("/configurations", configuration) ~> configurationController.route(accessToken) ~> check {
           status shouldBe StatusCodes.InternalServerError
           entityAs[String] shouldBe "Something went wrong"
@@ -52,10 +57,10 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
       val projectId = TestProjectUtils.getDummyProjectId
 
       "return configuration by existing project id" in {
-        when(configurationService.getById(projectId)).thenReturn(Future.successful(Some(configuration)))
+        when(configurationService.getById(projectId)).thenReturn(Future.successful(Some(configurationResponse)))
         Get("/configurations?project_id=" + projectId.value) ~> configurationController.route(accessToken) ~> check {
           status shouldBe StatusCodes.OK
-          entityAs[ProjectConfiguration] shouldBe configuration
+          entityAs[ProjectConfigurationEntity] shouldBe configurationResponse
         }
       }
 
@@ -64,6 +69,30 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
         Get("/configurations?project_id=" + projectId.value) ~> configurationController.route(accessToken) ~> check {
           status shouldBe StatusCodes.NotFound
           entityAs[String] shouldBe s"There is no configuration with project_id: ${projectId.value}"
+        }
+      }
+    }
+
+    "deactivate configuration" should {
+
+      "return success for deactivation configuration" in {
+        when(configurationService.deactivateConfiguration(configuration.projectId))
+          .thenReturn(Future.successful("Success"))
+        Delete("/configurations?project_id=" + configuration.projectId.value) ~> configurationController.route(
+          accessToken
+        ) ~> check {
+          status shouldBe StatusCodes.OK
+          entityAs[String] shouldBe "Success"
+        }
+      }
+
+      "return InternalServerError when failure deactivation configuration" in {
+        when(configurationService.deactivateConfiguration(configuration.projectId)).thenReturn(Future.failed(error))
+        Delete("/configurations?project_id=" + configuration.projectId.value) ~> configurationController.route(
+          accessToken
+        ) ~> check {
+          status shouldBe StatusCodes.InternalServerError
+          entityAs[String] shouldBe "Something went wrong"
         }
       }
     }
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala
@@ -4,6 +4,7 @@ import org.mongodb.scala.model.Filters._
 import org.mongodb.scala.model.ReplaceOptions
 import org.mongodb.scala.result.UpdateResult
 import org.mongodb.scala.{ Completed, Document, MongoCollection }
+import org.mongodb.scala.model.Updates._
 
 import scala.concurrent.Future
 
@@ -12,9 +13,12 @@ class DocumentRepository(collection: MongoCollection[Document]) {
   def addOne(document: Document): Future[Completed] =
     collection.insertOne(document).toFuture()
 
-  def updateOne(document: Document, fieldName: String, name: String): Future[UpdateResult] =
+  def replaceOne(document: Document, fieldName: String, name: String): Future[UpdateResult] =
     collection.replaceOne(equal(fieldName, name), document, ReplaceOptions().upsert(true)).toFuture()
 
   def getByParam(field: String, name: String): Future[Seq[Document]] =
-    collection.find(regex(field, name)).toFuture()
+    collection.find(equal(field, name)).toFuture()
+
+  def updateOneField(fieldName: String, fieldValue: String, updatedField: String, updatedValue: AnyVal): Future[UpdateResult] =
+    collection.updateOne(equal(fieldName, fieldValue), set(updatedField, updatedValue)).toFuture()
 }
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/TypedValue.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/TypedValue.scala
@@ -23,7 +23,7 @@ object TypedValue {
     implicit val floatTyped: OFormat[FloatTyped] = Json.format
     implicit val booleanTyped: OFormat[BooleanTyped] = Json.format
 
-    adtFormat("$type")(
+    adtFormat("_type")(
       adtCase[StringTyped]("String"),
       adtCase[FileTyped]("File"),
       adtCase[IntTyped]("Int"),
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
@@ -1,6 +1,7 @@
 package cromwell.pipeline.service
 
 import cromwell.pipeline.datastorage.dao.repository.DocumentRepository
+import cromwell.pipeline.datastorage.dto.project.configuration.ProjectConfigurationEntity
 import cromwell.pipeline.datastorage.dto.{ ProjectConfiguration, ProjectId }
 
 import scala.concurrent.{ ExecutionContext, Future }
@@ -8,11 +9,26 @@ import scala.concurrent.{ ExecutionContext, Future }
 class ProjectConfigurationService(repository: DocumentRepository)(implicit ec: ExecutionContext) {
   import ProjectConfiguration._
 
-  def addConfiguration(projectConfiguration: ProjectConfiguration): Future[String] =
+  def addConfiguration(configuration: ProjectConfigurationEntity): Future[String] =
     repository
-      .updateOne(toDocument(projectConfiguration), "projectId", projectConfiguration.projectId.value)
+      .replaceOne(
+        toDocument(ProjectConfiguration(configuration)),
+        "projectId",
+        configuration.projectId.value
+      )
       .map(_.toString)
 
-  def getById(projectId: ProjectId): Future[Option[ProjectConfiguration]] =
-    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument))
+  def getById(projectId: ProjectId): Future[Option[ProjectConfigurationEntity]] =
+    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).map {
+      case Some(conf) if conf.isActive =>
+        Some(ProjectConfigurationEntity(conf.projectId, conf.projectFileConfigurations))
+      case _ => None
+    }
+
+  def deactivateConfiguration(projectId: ProjectId): Future[String] =
+    getById(projectId).flatMap {
+      case Some(_) =>
+        repository.updateOneField("projectId", projectId.value, "isActive", false).map(_.toString)
+      case _ => Future.successful("There is no project to deactivate")
+    }
 }
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
@@ -3,6 +3,7 @@ package cromwell.pipeline.datastorage.dto
 import java.nio.file.Path
 
 import ProjectFile.pathFormat
+import cromwell.pipeline.datastorage.dto.project.configuration.ProjectConfigurationEntity
 import org.mongodb.scala.bson.collection.immutable.Document
 import play.api.libs.json.{ Json, OFormat }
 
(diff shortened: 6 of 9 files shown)


controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
```diff
@@ -27,9 +27,9 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
         ProjectFileConfiguration(Paths.get("/home/file"), List(FileParameter("nodeName", StringTyped(Some("hello")))))
       )
     )
+    val error = new RuntimeException("Something went wrong")
```
OstapStrashevskii, line 35: Why do you need to move this exception from 'should' block?

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
```diff
@@ -67,5 +67,29 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
         }
       }
     }
+
+    "deactivate configuration" should {
+
+      "return success for deactivation configuration" in {
+        when(configurationService.deactivateConfiguration(configuration.projectId))
+          .thenReturn(Future.successful("Success"))
+        Delete("/configurations?project_id=" + configuration.projectId.value) ~> configurationController.route(
```
OstapStrashevskii, line 81: Maybe it is easier to use concatenation here, but maybe, to hold one solid style you should rather change it to interpolation?

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
```diff
@@ -67,5 +67,29 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
...
+        }
+      }
+
+      "return InternalServerError when failure deactivation configuration" in {
+        when(configurationService.deactivateConfiguration(configuration.projectId)).thenReturn(Future.failed(error))
+        Delete("/configurations?project_id=" + configuration.projectId.value) ~> configurationController.route(
+          accessToken
+        ) ~> check {
+          status shouldBe StatusCodes.InternalServerError
+          entityAs[String] shouldBe "Something went wrong"
```
OstapStrashevskii, line 95: This sounds very abstract))  force your imagination and logic sense!) If I saw such an error I would feel scared)

OstapStrashevskii (changes requested):
Splendid! I wrote a few comments)

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
GrigoriiBerezin: I think, that we can leave as it is, because in this test we just test that the general error in service would be with expected behavior

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectConfigurationUtils.scala
```diff
@@ -0,0 +1,51 @@
...
+      inputsField <- Gen.nonEmptyListOf(fileParameterGen).suchThat(_.length < 4)
+    } yield ProjectFileConfiguration(pathField, inputsField)
+
+  lazy val projectConfigurationGen: Gen[ProjectConfiguration] =
+    for {
+      projectIdField <- projectIDGen
+      configurationsField <- Gen.nonEmptyListOf(projectFileConfigurationGen).suchThat(_.length < 4)
+    } yield ProjectConfiguration(projectIdField, configurationsField)
+
+  private lazy val stringGen: Gen[String] = Gen.alphaLowerStr.suchThat(str => str.length < 128 && str.nonEmpty)
```
GrigoriiBerezin: It's better to change `suchThat` onto `repeatUntil` because right now this gen can return None if it will be a string with higher length

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectConfigurationUtils.scala
```diff
@@ -0,0 +1,51 @@
...
+  lazy val fileParameterGen: Gen[FileParameter] =
+    for {
+      nameField <- stringGen
+      valueField <- typedValueGen
+    } yield FileParameter(nameField, valueField)
+
+  lazy val projectFileConfigurationGen: Gen[ProjectFileConfiguration] =
+    for {
+      pathField <- fileGen
+      inputsField <- Gen.nonEmptyListOf(fileParameterGen).suchThat(_.length < 4)
```
GrigoriiBerezin: change `nonEmptyListOf(fileParameterGen).suchThat(_.length < 4)` onto `listOfN(4, fileParameterGen)` for example to not get a large array

GrigoriiBerezin (changes requested):
Great work, there are some moment's to change also.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
Dmitrii-Shevchenko: to avoid duplication, I need it into "update" and "deactivate" tests

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
Dmitrii-Shevchenko: It's controversial, look at the UserControllerTest and ProjectControllerTest, we used the same style there. I think this is a good direction for a new task ;)

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectConfigurationUtils.scala
Dmitrii-Shevchenko: Maybe you mean retryUntil? Because it doesn't exist

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectConfigurationUtils.scala
Dmitrii-Shevchenko: done

OstapStrashevskii (approved):
Good work!

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectConfigurationUtils.scala
GrigoriiBerezin: Yep)

GrigoriiBerezin (approved):
Thanks you for your marvelous work!

myazinn:
I'd like to read the requirements to this task. What's its number in JIRA?

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala
```diff
@@ -16,5 +17,8 @@ class DocumentRepository(collection: MongoCollection[Document]) {
     collection.replaceOne(equal(fieldName, name), document, ReplaceOptions().upsert(true)).toFuture()
 
   def getByParam(field: String, name: String): Future[Seq[Document]] =
-    collection.find(regex(field, name)).toFuture()
+    collection.find(equal(field, name)).toFuture()
+
+  def deactivateOne(fieldName: String, name: String): Future[UpdateResult] =
+    collection.updateOne(equal(fieldName, name), set("active", "false")).toFuture()
```
myazinn: I'm not sure that the repository must know about the name of the field that we use as an activity field.
Maybe you should move all of this logic into `Service` layer. Read the Configuration, set activity to `false`, and then use `updateOne` method. This approach will also allow you to check that we are not trying to delete the configuration that is already removed.


Anyway, I assume activity is a boolean, not a string.
```suggestion
    collection.updateOne(equal(fieldName, name), set("active", false)).toFuture()
```

Dmitrii-Shevchenko:
@myazinn Added in description

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/TypedValue.scala
```diff
@@ -23,7 +23,7 @@ object TypedValue {
     implicit val floatTyped: OFormat[FloatTyped] = Json.format
     implicit val booleanTyped: OFormat[BooleanTyped] = Json.format
 
-    adtFormat("$type")(
+    adtFormat("_type")(
```
myazinn, line 26: Keep in mind that such a change will break deserialization if we already have some serialized classes.
Apparently we are not using it yet, but you should never do such things in a production project.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -15,4 +15,7 @@ class ProjectConfigurationService(repository: DocumentRepository)(implicit ec: E
 
   def getById(projectId: ProjectId): Future[Option[ProjectConfiguration]] =
     repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument))
```
myazinn: We should not let the user get `ProjectConfiguration` if it was deleted, aren't we?

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/project/configuration/ProjectConfigurationRequest.scala
```diff
@@ -0,0 +1,10 @@
+package cromwell.pipeline.datastorage.dto.project.configuration
+
+import cromwell.pipeline.datastorage.dto.{ProjectFileConfiguration, ProjectId}
+import play.api.libs.json.{Json, OFormat}
+
+case class ProjectConfigurationRequest(projectId: ProjectId, projectFileConfigurations: List[ProjectFileConfiguration])
+
+object ProjectConfigurationRequest {
+  implicit val projectConfigurationRequestFormat: OFormat[ProjectConfigurationRequest] = Json.format[ProjectConfigurationRequest]
+}
```
myazinn: Turn on `Ensure an empty line at the end of a file on Save` to get rid of these red warnings.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,33 @@
...
+        toDocument(ProjectConfiguration(configuration.projectId, configuration.projectFileConfigurations)),
+        "projectId",
+        configuration.projectId.value
+      )
       .map(_.toString)
 
-  def getById(projectId: ProjectId): Future[Option[ProjectConfiguration]] =
-    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument))
+  def getById(projectId: ProjectId): Future[Option[ProjectConfigurationResponse]] =
+    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).flatMap {
```
myazinn: Do we need `flatMap` here? I think `map` would be more appropriate.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Done

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala
Dmitrii-Shevchenko: Done

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/TypedValue.scala
Dmitrii-Shevchenko: It was @GrigoriiBerezin fix and I have not idea why there was $ before

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Like this?
``` repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).map(optionConf => optionConf.map(conf => if (conf.isActive) ProjectConfigurationResponse(conf.projectId, conf.projectFileConfigurations) else throw Exception))```

File: repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala (hunk 6 above)
Comment (myazinn) [reply repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala]: What about moving this logic to a `Service` layer or something like that?

File: services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala [hunk 11]
```diff
@@ -1,18 +1,33 @@
...
+      case Some(conf) if conf.isActive =>
+        Future.successful(Some(ProjectConfigurationResponse(conf.projectId, conf.projectFileConfigurations)))
+      case _ => Future.successful(None)
+    }
```
myazinn: @Dmitrii-Shevchenko I meant another thing
```suggestion
    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).map {
      case Some(conf) if conf.isActive =>
        Some(ProjectConfigurationResponse(conf.projectId, conf.projectFileConfigurations))
      case _ => None
    }
```

Dmitrii-Shevchenko: Looks like better, I didn't find this way, thanks!

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/TypedValue.scala
GrigoriiBerezin: We change start sign because Mongo doesn't accept fields with $ in the beginning.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,38 @@
...
-    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument))
+  def getById(projectId: ProjectId): Future[Option[ProjectConfigurationResponse]] =
+    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).map {
+      case Some(conf) if conf.isActive =>
+        Some(ProjectConfigurationResponse(conf.projectId, conf.projectFileConfigurations))
+      case _ => None
+    }
+
+  def deactivateConfiguration(projectId: ProjectId): Future[String] = {
+    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).flatMap {
```
myazinn: Reuse the `getById` method instead of duplicating its logic here.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,38 @@
...
+      case Some(conf) if conf.isActive =>
+        repository.updateOneField("projectId", projectId.value, "isActive", false).map(_.toString)
```
myazinn, line 31: Why not use`updateOne` method here? I don't think we want to keep creating methods like `updateTwoFields`, `updateThreeFields`, ets.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,38 @@
...
+      case _ => Future.successful("Activation field doesn't exist")
```
myazinn: How is this even possible that the `activation field` would not exist? :) What does it mean?

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: I thought update one field operation is faster then update whole document

myazinn: Maybe you are right, but I think for now the readability of the code is more important, than its performance. @ilyaonishenko what do you think?

myazinn: Forgot that Ilya is temporarily unavailable. @mnesvit what do you think? :)

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Done

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Done

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/repository/DocumentRepository.scala
Dmitrii-Shevchenko: Agree, already fixed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/project/configuration/ProjectConfigurationRequest.scala
Dmitrii-Shevchenko: Done

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Discussed it with @mnesvit and @ilyaonishenko at a meeting, decided that updateOneField method has right to exist ;)

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
myazinn: I don't see any changes. Same error message without explanation of how this field may not exists. Perhaps you meant something like `There is no project to deactivate`?

controllers/src/main/scala/cromwell/pipeline/controller/ProjectConfigurationController.scala
```diff
@@ -18,24 +19,30 @@ class ProjectConfigurationController(projectConfigurationService: ProjectConfigu
...
         get {
-          concat(
-            parameter('project_id.as[String]) { projectId =>
-              onComplete(projectConfigurationService.getById(ProjectId(projectId))) {
-                case Failure(e)                   => complete(StatusCodes.InternalServerError, e.getMessage)
-                case Success(Some(configuration)) => complete(configuration)
-                case Success(None) =>
-                  complete(StatusCodes.NotFound, s"There is no configuration with project_id: $projectId")
-              }
+          parameter('project_id.as[String]) { projectId =>
```
iliaonishchenko: It's deprecated way to specify parameters. Please don't use deprecated things. (and update Akka-http version)

portal/src/it/resources/application.conf
```diff
@@ -42,12 +42,19 @@ database {
...
+        host = "localhost"
+        host = ${?CMWL_MONGO_HOST}
+        port = "27017"
+        port = ${?CMWL_MONGO_PORT}
+        authenticationDatabase = "admin"
+        authenticationDatabase = ${?CMWL_MONGO_AUTH}
+        database = "mongo"
+        database = ${?CMWL_MONGO_DB}
+        collection = "configurations"
+        collection = ${?CMWL_MONGO_COLLECTION}
```
iliaonishchenko: Should we have such configuration in master branch?

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,37 @@
...
+        toDocument(ProjectConfiguration(configuration.projectId, configuration.projectFileConfigurations)),
+        "projectId",
+        configuration.projectId.value
+      )
       .map(_.toString)
 
-  def getById(projectId: ProjectId): Future[Option[ProjectConfiguration]] =
-    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument))
+  def getById(projectId: ProjectId): Future[Option[ProjectConfigurationResponse]] =
+    repository.getByParam("projectId", projectId.value).map(_.headOption.map(fromDocument)).map {
```
iliaonishchenko, line 22: If you're filtering,  use `filter`, not `map`.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,37 @@
...
+      case Some(conf) if conf.isActive =>
+        Some(ProjectConfigurationResponse(conf.projectId, conf.projectFileConfigurations))
+      case _ => None
+    }
+
+  def deactivateConfiguration(projectId: ProjectId): Future[String] =
```
iliaonishchenko, line 28: I'm not sure, is it a good thing that we have output result as a `String`? What other options do we have?

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,37 @@
...
+}
 import cromwell.pipeline.datastorage.dto.{ ProjectConfiguration, ProjectId }
 
 import scala.concurrent.{ ExecutionContext, Future }
 
 class ProjectConfigurationService(repository: DocumentRepository)(implicit ec: ExecutionContext) {
   import ProjectConfiguration._
 
-  def addConfiguration(projectConfiguration: ProjectConfiguration): Future[String] =
+  def addConfiguration(configuration: ProjectConfigurationRequest): Future[String] =
```
iliaonishchenko: The same question about `String`.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/project/configuration/ProjectConfigurationResponse.scala
```diff
@@ -0,0 +1,10 @@
+package cromwell.pipeline.datastorage.dto.project.configuration
+
+import cromwell.pipeline.datastorage.dto.{ProjectFileConfiguration, ProjectId}
+import play.api.libs.json.{Json, OFormat}
+
+final case class ProjectConfigurationResponse(projectId: ProjectId, projectFileConfigurations: List[ProjectFileConfiguration])
```
iliaonishchenko: I don't like the idea of having the same `ProjectConfigurationResponse` and `ProjectConfigurationRequest`. 
In this case more appropriate way is to create special single entity for the Mongo.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,18 +1,37 @@
...
     repository
-      .updateOne(toDocument(projectConfiguration), "projectId", projectConfiguration.projectId.value)
+      .replaceOne(
+        toDocument(ProjectConfiguration(configuration.projectId, configuration.projectFileConfigurations)),
```
iliaonishchenko: It's better to have an apply method in ProjectConfiguration companion object which takes `ProjectConfigurationResponse`.

iliaonishchenko (changes requested):
Thank you for your work!
There're several points to be fixed.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Repository returns ```Future[UpdateResult]```, is it better than ```String```?

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: ```Future[UpdateResult]```?

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: I'm not filtering, I have only one object from db, anyway, I would need to use ```map``` after ```filter``` like here:
```def getById(projectId: ProjectId): Future[Option[ProjectConfigurationEntity]] =
    repository
      .getByParam("projectId", projectId.value)
      .map(_.headOption.map(fromDocument))
      .filter {
        case Some(conf) => conf.isActive
        case None       => false
      }
      .map(conf => Some(ProjectConfigurationEntity(projectId, conf.get.projectFileConfigurations)))

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/project/configuration/ProjectConfigurationResponse.scala
Dmitrii-Shevchenko: United it into ProjectConfigurationEntity class

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Done

portal/src/it/resources/application.conf
Dmitrii-Shevchenko: Removed hardcoded lines

controllers/src/main/scala/cromwell/pipeline/controller/ProjectConfigurationController.scala
Dmitrii-Shevchenko: Сhanged according to last documentation. Version change entails a bunch of problems with HttpClient that does not apply to this task. But updating some libraries versions in dependencies is good idea for new one ;)

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
Dmitrii-Shevchenko: Yeah, you are right! Amended it on ```There is no project to deactivate```

myazinn (approved):
There are no more comments from me, thanks!
However, I agree with @ilyaonishenko that perhaps you should change the return type of the `deactivateConfiguration` method.

iliaonishchenko:
Hi, I'm sorry I cannot merge this because branch name is incorrect. Please create a MR from a correct branch and reference this MR.

iliaonishchenko:
@Dmitrii-Shevchenko I suppose this MR could be closed?

Dmitrii-Shevchenko:
> 
> 
> @Dmitrii-Shevchenko I suppose this MR could be closed?

Yep

PR #85, opened 2020-07-17
EPMLSTRCMW-165: add integration tests for projectconfiguration

IaroslavTavchenkov:


portal/src/it/scala/cromwell/pipeline/controller/ProjectConfigurationControllerItTest.scala
```diff
@@ -0,0 +1,46 @@
...
+    "name" -> "MongoDB",
+    "type" -> "database",
+    "count" -> 1,
+    "info" -> Document("x" -> 203, "y" -> 102)
+  )
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+    components.datastorageModule.pipelineDatabaseEngine.updateSchema()
+    val x = configurationRepository.addOne(doc)
```
IaroslavTavchenkov, line 45: What should I do with this Type Mismatch?

myazinn: I think `val doc = Document(` instead of `val doc = DBO(` will solve the problem

IaroslavTavchenkov: > I think `val doc = Document(` instead of `val doc = DBO(` will solve the problem

Yeah, but the idea was to use this library for MongoDB instead of Document()
https://index.scala-lang.org/osinka/subset2/subset/2.2.3?target=_2.12

'cuz our json can be really big.

myazinn: @IaroslavTavchenkov Sorry, I've missed this answer :( Is your question still relevant?

PR #86, opened 2020-07-27
feat: docker-compose for postgres and mongo

Dmitrii-Shevchenko:
EPMLSTRCMW-168
Implementation non-official docker compose file

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,5 @@
-version: "3"
+version: '3'
 services:
-  #  Create a service named db.
-  db:
-    image: "postgres:12-alpine"
-    container_name: "pipeline_db"
-    #   Setup the username, password, and database name
-    environment:
-      - POSTGRES_USER=pipeline
-      - POSTGRES_PASSWORD=pipeline
-      - POSTGRES_DB=pipeline
-    #   Map port 6432 (localhost) to port 5432 on the container
-    ports:
-      - "6432:5432"
-
   gitlab:
     image: 'gitlab/gitlab-ce:latest'
     restart: always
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,31 +1,40 @@
...
+#       nginx['listen_port'] = 9080                   # make nginx to listen on the same port as confgured in external_url
+#       # Add any other gitlab.rb configuration here, each on its own line
+#    ports:
+#      - '9080:9080'
+#      - '9443:443'
+#      - '9022:22'
+#    volumes:
+#      - '/srv/gitlab/config:/etc/gitlab'
+#      - '/srv/gitlab/logs:/var/log/gitlab'
+#      - '/srv/gitlab/data:/var/opt/gitlab'
```
OstapStrashevskii: Why do you need it here? Maybe we need some explanation in Readme or todo comment?

OstapStrashevskii (changes requested):
Good work! )

docker-compose.yml
GrigoriiBerezin: Agree with Ostap

GrigoriiBerezin (approved):
Nicely done

docker-compose.yml
Dmitrii-Shevchenko: @OstapStrashevskii @GrigoriiBerezin 
because it`s non-official docker compose file was been created for the convenience of developers
but you can discuss it with @ilyaonishenko 
https://jirapct.epam.com/jira/browse/EPMLSTRCMW-168

Dmitrii-Shevchenko: done

OstapStrashevskii (approved):
Nice done!

IaroslavTavchenkov (approved):
Good Job!

mnesvit (commented):
Should it work only in tests or we are planning to run application with this docker files? In case of second option it's necessary to mount volumes for data.

Dmitrii-Shevchenko:
@mnesvit So far this is only for tests, but in the next task I will mount volumes

mnesvit (approved):

iliaonishchenko (approved):
It's beautiful!

PR #92, opened 2020-08-18
Add api collections for import to postman

GrigoriiBerezin:


iliaonishchenko (changes requested):
It would be nice for you to be aligned with our style guide.
For example: branch naming, commit messages. 
If there is no such ticket in Jira, you could create one or use `EPMLSTRCMW-00` code.

PR #97, opened 2020-08-25
fix: project post request (add project)

Dmitrii-Shevchenko:
EPMLSTRCMW-173

ProjectController's POST method pattern matching converted all Success futures to OK status, when it should be divided to Left and Right conditions. Fixed.

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
@@ -25,7 +25,7 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
     }
   }
 
-  def addProject(request: ProjectAdditionRequest, userId: UserId): Future[Either[VersioningException, ProjectId]] = {
+  def addProject(request: ProjectAdditionRequest, userId: UserId): Future[ProjectId] = {
     val project =
       Project(
         projectId = ProjectId(UUID.randomUUID().toString),
@@ -35,7 +35,7 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
       )
     projectVersioning.createRepository(project).flatMap {
       case Left(exception)       => Future.failed(exception)
-      case Right(value: Project) => projectRepository.addProject(value).map(Right(_))
+      case Right(value: Project) => projectRepository.addProject(value)
     }
   }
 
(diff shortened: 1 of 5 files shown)


OstapStrashevskii (approved):
The nice and actual request!

IaroslavTavchenkov (approved):
Good job! Thanks for help!)

services/src/main/scala/cromwell/pipeline/service/VersioningException.scala
```diff
@@ -1,13 +1,16 @@
 package cromwell.pipeline.service
 
-sealed abstract class VersioningException extends Exception {val message: String}
+sealed abstract class VersioningException extends Exception {
+  val message: String
+  override def getMessage() = message
+}
```
myazinn: There is another way to make it
```suggestion
sealed abstract class VersioningException(message: String) extends Exception(message)
```
However, it will force you to write
`case class ProjectException(message: String) extends VersioningException(message)`
instead of
`case class ProjectException(message: String) extends VersioningException()`
Nevertheless, in my opinion, it is a little bit better.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -34,7 +34,7 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
         active = true
       )
     projectVersioning.createRepository(project).flatMap {
-      case Left(exception)       => Future.successful(Left(exception))
+      case Left(exception)       => Future.failed(exception)
```
myazinn: What is the point of returning `Either[A, B]` if we always return Right?
What problem you are trying to solve with this fix?

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/VersioningException.scala
Dmitrii-Shevchenko: Agree, it looks like better

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
Dmitrii-Shevchenko: Changed returning on ```Future[ProjectId]```. The problem is we get Left object as a success condition in the controller, but it has more to do with failure

myazinn: Maybe you should have fix the controller? :)
It can return some custom response code in case of `VersioningException`

Dmitrii-Shevchenko: Changed on matching in controller

myazinn (approved):

iliaonishchenko (approved):

PR #99, opened 2020-08-31
EPMLSTRCMW-172: incorrect response error

OstapStrashevskii:
Log in with email of <User> and <Wrong Password>,
<User> get "Wrong password" response

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/AuthService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/AuthService.scala
@@ -7,15 +7,10 @@ import cats.implicits._
 import cromwell.pipeline.auth.AuthUtils
 import cromwell.pipeline.datastorage.dao.repository.UserRepository
 import cromwell.pipeline.datastorage.dto.User
-import cromwell.pipeline.datastorage.dto.auth.{
-  AccessTokenContent,
-  AuthContent,
-  AuthResponse,
-  RefreshTokenContent,
-  SignInRequest,
-  SignUpRequest
-}
+import cromwell.pipeline.datastorage.dto.auth._
 import cromwell.pipeline.model.wrapper.UserId
+import cromwell.pipeline.service.AuthService.authorizationFailure
+import cromwell.pipeline.service.AuthorizationException.IncorrectPasswordException
 import cromwell.pipeline.utils.StringUtils
 import play.api.libs.json.Json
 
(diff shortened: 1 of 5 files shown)


IaroslavTavchenkov (approved):
Seems very good!)

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -26,16 +20,22 @@ class AuthService(userRepository: UserRepository, authUtils: AuthUtils)(implicit
...
-        val accessTokenContent = AccessTokenContent(user.userId)
-        val refreshTokenContent = RefreshTokenContent(user.userId, None)
-        getAuthResponse(accessTokenContent, refreshTokenContent, Instant.now.getEpochSecond)
+  def signIn(request: SignInRequest): Future[Option[AuthResponse]] = responseFromOptUser(authFilter(request))
+
+  def authFilter(request: SignInRequest): OptionT[Future, User] =
+    OptionT(userRepository.getUserByEmail(request.email)).semiflatMap[User] { user =>
+      passwordCorrect(request, user) match {
+        case true  => Future(user)
+        case false => Future.failed(IncorrectPasswordException("Wrong password"))
```
Dmitrii-Shevchenko: In my opinion, if-else is more appropriate here. Base on [it](https://stackoverflow.com/questions/9266822/pattern-matching-vs-if-else)

OstapStrashevskii: Nice advice! done! )

Dmitrii-Shevchenko (approved):
Good work!

GrigoriiBerezin (approved):
Nicely done

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -26,16 +20,22 @@ class AuthService(userRepository: UserRepository, authUtils: AuthUtils)(implicit
...
+    OptionT(userRepository.getUserByEmail(request.email)).semiflatMap[User] { user =>
+      if (passwordCorrect(request, user)) Future(user)
+      else {
+        Future.failed(IncorrectPasswordException("Wrong password"))
       }
-      .value
-      .map(_.flatten)
+    }
+
+  def responseFromOptUser(optUser: OptionT[Future, User]): Future[Option[AuthResponse]] =
```
myazinn: It may be nitpicking, but this code doesn't look like Scala, despite the fact that you are using cats here.
We love `.map`, `flatMap` etc, since they allow us to abstract over the type of the container (such as `Option`, `OptionT`) and use one function with multiple containers without any problem. You can't reuse your `def responseFromOptUser` with any other type of container (even if it is a simple `Option`).
It would be better to change the `responseFromOptUser` method as follows.
`def responseFromUser(user: User): Option[AuthResponse]`
This method shouldn't care about addtional wrappers like `OptionT`. If you give him a user, he will give a response. That's all.
In that case, `signIn` method should be rewritten as follows
```
def signIn(request: SignInRequest): Future[Option[AuthResponse]] =
  authFilter(request).subflatMap(responseFromUser).value
```
Also, the name of the method `authFilter`  does not correspond to what it does. It takes `SignInRequest` and returns `User`. According to the name, it should only filter something. According to the types, it somehow transforms `request` into `user`.
And don't forget that method names should be verbs.

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -26,16 +20,22 @@ class AuthService(userRepository: UserRepository, authUtils: AuthUtils)(implicit
...
-      .filter(user => user.passwordHash == StringUtils.calculatePasswordHash(request.password.value, user.passwordSalt))
-      .map { user =>
-        val accessTokenContent = AccessTokenContent(user.userId)
-        val refreshTokenContent = RefreshTokenContent(user.userId, None)
-        getAuthResponse(accessTokenContent, refreshTokenContent, Instant.now.getEpochSecond)
+  def signIn(request: SignInRequest): Future[Option[AuthResponse]] = responseFromOptUser(authFilter(request))
+
+  def authFilter(request: SignInRequest): OptionT[Future, User] =
+    OptionT(userRepository.getUserByEmail(request.email)).semiflatMap[User] { user =>
+      if (passwordCorrect(request, user)) Future(user)
```
myazinn:
```suggestion
      if (passwordCorrect(request, user)) Future.successful(user)
```

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
OstapStrashevskii: 1) "nitpicking", - no, it is not the case!
2) your mention is pretty cool, I suppose I just need more cats exp to observe all possible variants!

Done

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
OstapStrashevskii: done! X)

myazinn (approved):

iliaonishchenko (approved):
Nice job!
Looks good to me!

PR #105, opened 2020-10-14
EPMLSTRCMW-185: Print all configuration parameters during the application initialization

Dmitrii-Shevchenko:
Print all configuration parameters during the application initialization
Example logs:
```
{
  "auth" : {
    "secretKey" : "****************",
    "hmacAlgorithm" : "HS256",
    "expirationTimeInSeconds" : {
      "accessToken" : 300,
      "refreshToken" : 900,
      "userSession" : 3600
    }
  },
  "webservice" : {
    "interface" : "localhost",
    "port" : 8080
  },
  "mongo" : {
    "user" : "mongoUser",
    "password" : "********",
    "host" : "localhost",
    "port" : 27017,
    "authenticationDatabase" : "admin",
    "database" : "mongo",
    "collection" : "configurations"
  },
  "postgres" : {
    "user" : "postgres",
    "password" : "******",
    "host" : "localhost",
    "port" : 5432,
    "database" : "postgres"
  },
  "gitlab" : {
    "url" : "https://gitlab.com/api/v4/",
    "idPath" : "root%2F",
    "token" : {
      "PRIVATE-TOKEN" : "********************"
    },
    "defaultFileVersion" : "v0.0.1",
    "defaultBranch" : "master"
  }
}
```

### DIFF
--- a/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
+++ b/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
@@ -8,7 +8,9 @@ import akka.http.scaladsl.server.{ RejectionHandler, Route, ValidationRejection 
 import akka.stream.ActorMaterializer
 import cromwell.pipeline.auth.token.MissingAccessTokenRejection
 import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.utils.ApplicationConfig
 import org.slf4j.LoggerFactory
+import play.api.libs.json._
 
 import scala.concurrent.ExecutionContextExecutor
 
@@ -20,6 +22,18 @@ object CromwellPipelineApp extends App {
 
   val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
   val components = new ApplicationComponents()
+
+  log.info(
+    Json.prettyPrint(
+      Json.obj(
+        "auth" -> Json.toJson(ApplicationConfig.load().authConfig),
+        "webservice" -> Json.toJson(ApplicationConfig.load().webServiceConfig),
+        "mongo" -> Json.toJson(ApplicationConfig.load().mongoConfig),
+        "postgres" -> Json.toJson(ApplicationConfig.load().postgresConfig),
+        "gitlab" -> Json.toJson(ApplicationConfig.load().gitLabConfig)
+      )
+    )
+  )
 
   import components.applicationConfig.webServiceConfig
   import components.controllerModule._
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -22,7 +22,7 @@ object Dependencies {
     val liquibase = "3.8.6"
     val postgresql = "42.2.8"
     val womtool = "48"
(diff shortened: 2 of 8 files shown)


project/Dependencies.scala
```diff
@@ -22,7 +22,7 @@ object Dependencies {
     val liquibase = "3.8.6"
     val postgresql = "42.2.8"
     val womtool = "48"
-    val logback = "1.2.3"
+    val logback = "1.3.0-alpha5"
```
myazinn, line 25: It is better to use stable library versions. Why do you want to use unstable one? It doesn't look like you are using any new features.

utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala
```diff
@@ -3,8 +3,64 @@ package cromwell.pipeline.utils
...
 import pdi.jwt.algorithms.JwtHmacAlgorithm
+import play.api.libs.json.{ JsObject, Json, Writes }
 
 sealed trait ConfigComponent
+object ConfigComponent {
+  implicit val webserviceWrites = new Writes[WebServiceConfig] {
+    def writes(config: WebServiceConfig) = Json.obj(
+      "interface" -> config.interface,
+      "port" -> config.port
+    )
```
myazinn, line 14: Use typesafe json combinators instead. (https://www.playframework.com/documentation/2.8.x/ScalaJsonCombinators#Writes).
If someone will add a new field to a config, your implementation of `Writes` will compile. Sometimes it really is what you need, but in this case it is not. Using a combinator will ensure that if someone adds a field to the config, the code will not compile until he fixes `Writer`.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -21,6 +23,18 @@ object CromwellPipelineApp extends App {
...
   val components = new ApplicationComponents()
 
+  log.info(
+    Json.prettyPrint(
+      Json.obj(
+        "auth" -> Json.toJson(ApplicationConfig.load().authConfig),
+        "webservice" -> Json.toJson(ApplicationConfig.load().webServiceConfig),
+        "mongo" -> Json.toJson(ApplicationConfig.load().mongoConfig),
+        "postgres" -> Json.toJson(ApplicationConfig.load().postgresConfig),
+        "gitlab" -> Json.toJson(ApplicationConfig.load().gitLabConfig)
```
myazinn, line 33: 1) No need to read whole Application config every time from scratch since all of them are already available in `component` variable.
2) Putting so much logic about config printing inside initialization object is wrong. For example, you can create something like `ConfigRenderer` object in utils, that will have only one method
`def render(configs: ConfigComponent*): String`
And then rewrite line 26 as
```
import components.applicationConfig._
log.info(ConfigRenderer.render(authConfig, webServiceConfig, mongoConfig, postgresConfig, gitLabConfig))
```

myazinn (changes requested):

Dmitrii-Shevchenko:
Expired

PR #106, opened 2020-10-15
EPMLSTRCMW-184

Phoenix100101:


portal/src/it/resources/application.conf
```diff
@@ -51,3 +51,9 @@ database {
     collection = ${?CMWL_TEST_MONGO_COLLECTION}
   }
 }
+
+cromwell-backend {
+    host = ${?CMWL_AWS_HOST}
```
mnesvit: Why are we using "AWS" segment here? It could be tricky. Maybe it would be better to take "CROMWELL"? Doesn't sound really good with prefix, but more obvious.

portal/src/main/resources/application.conf
```diff
@@ -58,3 +58,9 @@ database {
     collection = ${?CMWL_MONGO_COLLECTION}
   }
 }
+
+cromwell-backend {
+    host = ${?CMWL_AWS_HOST}
```
mnesvit: Same question with "AWS" segment here

controllers/src/main/scala/cromwell/pipeline/controller/CromwellController.scala
```diff
@@ -0,0 +1,30 @@
...
+class CromwellController(cromwellBackendService: CromwellService)(
+  implicit executionContext: ExecutionContext
+) {
+
+  val route: AccessTokenContent => Route = _ => {
+    path("cromwell" / "engine" / "status") {
+      concat(
+        get {
+          onComplete(cromwellBackendService.getEngineStatus()) {
+            case Success(Right(state)) => complete(StatusCodes.OK, state.toString)
```
myazinn: `.toString` is not required, `complete(StatusCodes.OK, state)` should work just fine.

services/src/main/scala/cromwell/pipeline/service/CromwellService.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import cromwell.pipeline.utils.{CromwellConfig, HttpStatusCodes}
+import play.api.libs.json.JsObject
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class CromwellService(httpClient: HttpClient, cromwellConfig: CromwellConfig)(
+  implicit executionContext: ExecutionContext
+) {
+  def getEngineStatus(): Future[Either[Exception, Int]] = {
```
myazinn:
```suggestion
  def getEngineStatus(): Future[Either[CromwellIntegrationException, Int]] = {
```
The whole point of using `Future[Either[_, _]]` instead of simple `Future[_]` is to allow us to narrow types of the expected errors. In fact, `Future[Either[Exception, _]]` almost as useless (in terms of type information) as `Future[_]`.

services/src/main/scala/cromwell/pipeline/service/CromwellService.scala
```diff
@@ -0,0 +1,22 @@
...
+    val awsUrl: String = cromwellConfig.host + cromwellConfig.enginePath + "/" + cromwellConfig.version + "/status"
```
myazinn: In Scala, we prefer string interpolation over string concatenation. E.g.
`val string = s"$a/$b/status"`
rather than
`val string = a + "/" + b + "/status"`
At a first glance, it may look cryptic and weird, but it might save from some unexpected bugs (e.g. if `cromwellConfig.host` and `cromwellConfig.enginePath` were `Int`s).
Also, I think this can be class field instead of local variable.

myazinn (changes requested):
Nice work! Here are some comments though :)

myazinn (approved):
Nice work!

mnesvit (approved):

iliaonishchenko:
Travis failed. 
We also have a style guide how to write commit messages and branch names. 
Could you fix your commit messages please?
Also, branch naming is not appropriate here, but that's ok for current MR. But please be aware in the future. (It probably my fault we don't have branch name conventions in source code)
EPMLSTRCMW-184_Add_health_check_method_for_cromwell -- is the correct way to do it.

Also, it would be nice if you write a MR description for us to understand what is happening in this MR. And very small description in the MR name. (Similar to the branch name)

iliaonishchenko:
Closing because of the inactivity

PR #108, opened 2020-10-21
Epmlstrcmwl 90 run dao

manvicha:
Enabling the run DAO to access to the postgres table,
basic methods were added add run, update run, get run
and delete run.

### DIFF
--- a/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
+++ b/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
@@ -65,4 +65,25 @@
         </rollback>
     </changeSet>
 
+    <changeSet id="create_run_table" author="manuel_vigueras">
+        <createTable tableName="run">
+            <column name="run_id" type="${uuid.type}"/>
+            <column name="project_id" type="${uuid.type}"/>
+            <column name="project_version" type="${text.type}"/>
+            <column name="status" type="${text.type}"/>
+            <column name="time_start" type="${timestamp.type}"/>
+            <column name="time_end" type="${timestamp.type}"/>
+            <column name="user_id" type="${uuid.type}"/>
+            <column name="results" type="${text.type}"/>
+            <column name="cmwl_workflow_id" type="${text.type}"/>
+        </createTable>
+        <addPrimaryKey constraintName="run_pk" tableName="run" columnNames="run_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_user" baseTableName="run" baseColumnNames="user_id" referencedTableName="user" referencedColumnNames="user_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_project" baseTableName="run" baseColumnNames="project_id" referencedTableName="project" referencedColumnNames="project_id"/>
+        <addNotNullConstraint constraintName="project_version_nn" tableName="run" columnName="project_version"/>
+        <addNotNullConstraint constraintName="status_nn" tableName="run" columnName="status"/>
+        <addNotNullConstraint constraintName="time_start_nn" tableName="run" columnName="time_start"/>
+        <addNotNullConstraint constraintName="results_nn" tableName="run" columnName="results"/>
+    </changeSet>
+
 </databaseChangeLog>
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
@@ -1,11 +1,17 @@
 package cromwell.pipeline.datastorage
 
 import cromwell.pipeline.database.{ MongoEngine, PipelineDatabaseEngine }
-import cromwell.pipeline.datastorage.dao.entry.UserEntry
-import cromwell.pipeline.datastorage.dao.repository.{ DocumentRepository, ProjectRepository, UserRepository }
-import cromwell.pipeline.datastorage.dao.{ ProjectEntry, ProjectProfileWithEnumSupport }
+import cromwell.pipeline.datastorage.dao.ProjectEntry
+import cromwell.pipeline.datastorage.dao.entry.{ RunEntry, UserEntry }
+import cromwell.pipeline.datastorage.dao.repository.{
+  DocumentRepository,
+  ProjectRepository,
+  RunRepository,
+  UserRepository
+}
+import cromwell.pipeline.datastorage.dto.CustomsWithEnumSupport
 import cromwell.pipeline.model.validator.{ Enable, Wrapped }
-import cromwell.pipeline.model.wrapper.{ Name, UserEmail, UserId }
+import cromwell.pipeline.model.wrapper.{ Name, RunId, UserEmail, UserId }
 import cromwell.pipeline.utils.ApplicationConfig
 import org.mongodb.scala.{ Document, MongoCollection }
 import slick.jdbc.JdbcProfile
@@ -23,6 +29,8 @@ class DatastorageModule(applicationConfig: ApplicationConfig) {
     new UserRepository(pipelineDatabaseEngine, databaseLayer)
   lazy val projectRepository: ProjectRepository =
     new ProjectRepository(pipelineDatabaseEngine, databaseLayer)
+  lazy val runRepository: RunRepository =
+    new RunRepository(pipelineDatabaseEngine, databaseLayer)
   lazy val configurationRepository: DocumentRepository = new DocumentRepository(configurationCollection)
 }
 
@@ -36,6 +44,7 @@ trait Profile {
     import scala.language.implicitConversions
 
     implicit def uuidIso: Isomorphism[UserId, String] = iso[UserId, String](_.unwrap, UserId(_, Enable.Unsafe))
+    implicit def runidIso: Isomorphism[RunId, String] = iso[RunId, String](_.unwrap, RunId(_, Enable.Unsafe))
     implicit def emailIso: Isomorphism[UserEmail, String] =
       iso[UserEmail, String](_.unwrap, UserEmail(_, Enable.Unsafe))
     implicit def nameIso: Isomorphism[Name, String] = iso[Name, String](_.unwrap, Name(_, Enable.Unsafe))
@@ -51,4 +60,5 @@ class DatabaseLayer(override val profile: JdbcProfile)
     extends Profile
     with UserEntry
     with ProjectEntry
-    with ProjectProfileWithEnumSupport
+    with CustomsWithEnumSupport
+    with RunEntry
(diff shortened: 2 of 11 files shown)


manvicha:
I have a doubt about the delete method, I saw that you are implementing logic deletes but not sure if for this table should be the same. In case yes, what would be the column?

datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -65,4 +65,21 @@
...
+            <column name="time_start" type="${timestamp.type}"/>
+            <column name="time_end" type="${timestamp.type}"/>
+            <column name="user_id" type="${uuid.type}"/>
+            <column name="results" type="${text.type}"/>
+            <column name="cmwl_workflow_id" type="${text.type}"  defaultValueBoolean="true"/>
+        </createTable>
+        <addPrimaryKey constraintName="run_pk" tableName="run" columnNames="run_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_user" baseTableName="run" baseColumnNames="user_id" referencedTableName="user" referencedColumnNames="user_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_project" baseTableName="run" baseColumnNames="project_id" referencedTableName="project" referencedColumnNames="project_id"/>
+    </changeSet>
```
myazinn, line 87: You should also add `not null` constraints for every field that you are consider to be required. In the `Run` class all fields are required (there are no `Option[_]` fields for now), which means this constraint must be added for all of these fields.

repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
```diff
@@ -1,13 +1,13 @@
...
-import cromwell.pipeline.datastorage.dao.repository.{ DocumentRepository, ProjectRepository, UserRepository }
-import cromwell.pipeline.datastorage.dao.{ ProjectEntry, ProjectProfileWithEnumSupport }
-import cromwell.pipeline.model.validator.{ Enable, Wrapped }
-import cromwell.pipeline.model.wrapper.{ Name, UserEmail, UserId }
+import cromwell.pipeline.database.{MongoEngine, PipelineDatabaseEngine}
+import cromwell.pipeline.datastorage.dao.entry.{RunEntry, UserEntry}
+import cromwell.pipeline.datastorage.dao.repository.{DocumentRepository, ProjectRepository, RunRepository, UserRepository}
+import cromwell.pipeline.datastorage.dao.{ProjectEntry, ProjectProfileWithEnumSupport}
+import cromwell.pipeline.model.validator.{Enable, Wrapped}
+import cromwell.pipeline.model.wrapper.{Name, RunId, UserEmail, UserId}
```
myazinn: Have you configured `scalafmt`?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
```diff
@@ -0,0 +1,34 @@
+package cromwell.pipeline.datastorage.dto
+
+import java.sql.Timestamp
```
myazinn: I think we should use `Instant` or `OffsetDateTime` instead of `Timestamp` to abstract over `sql`.
I guess `Run` class will be used in Services and Controllers, which must never "know" that we are using `SQL` somewhere.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import cromwell.pipeline.model.wrapper.{RunId, UserId}
+import play.api.libs.json.{Format, JsResult, JsValue, Json, OFormat}
+
+final case class Run(
+  runId: RunId,
+  projectId: ProjectId,
+  projectVersion: String,
+  status: String,
```
myazinn: Use ADT instead of `String`. For example, there's a `Visibility` trait that solves the same problem.
@ilyaonishenko Maybe it's time to use some suitable library, like `enumeratum`?

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/RunRepositoryTest.scala
```diff
@@ -0,0 +1,77 @@
...
+            status = "new-status",
+            timeStart = TestRunUtils.getDummyTimestamp,
+            timeEnd = TestRunUtils.getDummyTimestamp,
+            results = "new-result",
+            cmwlWorkflowId = "new-cmwlWorkflowId"
+          )
+
+        val result = for {
+          _ <- runRepository.updateRun(updatedRun)
+          getById <- runRepository.getRunById(dummyRun.runId)
```
myazinn, line 63: In this test, you are not inserting new `Run`s. What are you trying to update?
It is considered bad practice when your test depend on other test. If you reorder your tests, this test will fail, since there are no `runs` in DB and you will get `None` instead of `Some(updatedRun)`

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/RunRepositoryTest.scala
```diff
@@ -0,0 +1,77 @@
...
+        result.map(optRun => optRun shouldEqual Some(updatedRun))
+      }
+    }
+
+    "delete run" should {
+
+      "delete run by id" taggedAs Dao in {
+        val result = for {
+          _ <- runRepository.deleteRunById(dummyRun.runId)
+          getById <- runRepository.getRunById(dummyRun.runId)
```
myazinn, line 82: Same problem as with `update`.
You never inserted `Run` in this test, therefore this test should always pass even if `deleteRunById` does nothing.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
```diff
@@ -0,0 +1,34 @@
...
+  timeStart: Timestamp,
+  timeEnd: Timestamp,
```
myazinn: What `timeEnd` should `Run` have if it is not yet completed?

myazinn (changes requested):

datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
manvicha: Changed it. constraints were added

repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
manvicha: scalafmt applied

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
manvicha: Changed it to use Instant

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
manvicha: I don't get it, why use Visibility? Is there something that we need to wrap it?

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/RunRepositoryTest.scala
manvicha: Thanks for the comments, it makes a lot of sense. I didn't think about what would happen if we reorder the tests, changes were applied

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
manvicha: Changed to Option[Instant]

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Run.scala
myazinn: You shouldn't use `Visibility` at all. It's just an example of how to introduce ADT and integrate it with Slick.

manvicha: I created a new dto called ```CustomsWithEnumSupport```. In this file i placed the ```Visibility``` and ```Status```

datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -65,4 +65,26 @@
...
+            <column name="cmwl_workflow_id" type="${text.type}"  defaultValueBoolean="true"/>
+        </createTable>
+        <addPrimaryKey constraintName="run_pk" tableName="run" columnNames="run_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_user" baseTableName="run" baseColumnNames="user_id" referencedTableName="user" referencedColumnNames="user_id"/>
+        <addForeignKeyConstraint constraintName="fk_run_project" baseTableName="run" baseColumnNames="project_id" referencedTableName="project" referencedColumnNames="project_id"/>
+        <addNotNullConstraint constraintName="project_version_nn" tableName="run" columnName="project_version"/>
+        <addNotNullConstraint constraintName="status_nn" tableName="run" columnName="status"/>
+        <addNotNullConstraint constraintName="time_start_nn" tableName="run" columnName="time_start"/>
+        <addNotNullConstraint constraintName="results_nn" tableName="run" columnName="results"/>
+        <addNotNullConstraint constraintName="cmwl_workflow_id_nn" tableName="run" columnName="cmwl_workflow_id"/>
```
mnesvit: Cromwell workflow id should be nullable (in code and database)

manvicha: Change applied it

datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -65,4 +65,25 @@
...
+        <createTable tableName="run">
+            <column name="run_id" type="${uuid.type}"/>
+            <column name="project_id" type="${uuid.type}"/>
+            <column name="project_version" type="${text.type}"/>
+            <column name="status" type="${text.type}"/>
+            <column name="time_start" type="${timestamp.type}"/>
+            <column name="time_end" type="${timestamp.type}"/>
+            <column name="user_id" type="${uuid.type}"/>
+            <column name="results" type="${text.type}"/>
+            <column name="cmwl_workflow_id" type="${text.type}"  defaultValueBoolean="true"/>
```
myazinn: Better to remove this strange default value.
```suggestion
            <column name="cmwl_workflow_id" type="${text.type}"/>
```

Sorry, I just noticed this.

manvicha: Awesome catch! Thanks Nikita! Changed it

myazinn (approved):

mnesvit (approved):

PR #115, opened 2020-11-02
feat: Implement Run Controller

manvicha:
Enabling the run controller to access to the run services,
basic methods were added add run, update run, get run
and delete run.

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/RunController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/RunController.scala
@@ -1,0 +1,60 @@
+package cromwell.pipeline.controller
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives.{ entity, _ }
+import akka.http.scaladsl.server.Route
+import cromwell.pipeline.controller.utils.FromStringUnmarshallers._
+import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.datastorage.dto.{ RunCreateRequest, RunDeleteRequest, RunUpdateRequest }
+import cromwell.pipeline.model.wrapper.RunId
+import cromwell.pipeline.service.RunService
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+
+import scala.concurrent.ExecutionContext
+import scala.util.{ Failure, Success }
+
+class RunController(runService: RunService)(
+  implicit executionContext: ExecutionContext
+) {
+
+  val route: AccessTokenContent => Route = accessToken =>
+    path("runs") {
+      concat(
+        get {
+          parameter('run_id.as[RunId]) { runId =>
+            {
+              onComplete(runService.getRunByIdAndUser(runId, accessToken.userId)) {
+                case Success(Some(response)) => complete(response)
+                case Success(None)           => complete(StatusCodes.NotFound, "Run not found")
+                case Failure(_)              => complete(StatusCodes.InternalServerError, "Internal error")
+              }
+            }
+          }
+        },
+        delete {
+          entity(as[RunDeleteRequest]) { request =>
+            onComplete(runService.deleteRunById(request.runId, accessToken.userId)) {
+              case Success(idResponse) => complete(idResponse)
+              case Failure(_)          => complete(StatusCodes.InternalServerError, "Internal error")
+            }
+          }
+        },
+        put {
+          entity(as[RunUpdateRequest]) { runUpdateRequest =>
+            onComplete(runService.updateRun(runUpdateRequest, accessToken.userId)) {
+              case Success(_)   => complete(StatusCodes.NoContent)
+              case Failure(exc) => complete(StatusCodes.InternalServerError, exc.getMessage)
+            }
+          }
+        },
+        post {
+          entity(as[RunCreateRequest]) { request =>
+            onComplete(runService.addRun(request)) {
+              case Success(response) => complete(response)
+              case Failure(_)        => complete(StatusCodes.InternalServerError, "Internal error")
+            }
+          }
+        }
+      )
+    }
+}
--- a/controllers/src/test/scala/cromwell/pipeline/controller/RunControllerTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/RunControllerTest.scala
@@ -1,0 +1,189 @@
+package cromwell.pipeline.controller
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.MalformedQueryParamRejection
+import akka.http.scaladsl.testkit.ScalatestRouteTest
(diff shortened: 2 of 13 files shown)


controllers/src/test/scala/cromwell/pipeline/controller/RunControllerTest.scala
```diff
@@ -0,0 +1,178 @@
...
+        val dummyRun: Run = TestRunUtils.getDummyRun()
+        val runId = dummyRun.runId
+        val runRespOption: Option[Run] = Option(dummyRun)
+
+        when(runService.getRunById(runId)).thenReturn(Future.successful(runRespOption))
+
+        Get("/runs?run_id=" + runId.unwrap) ~> runController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.OK
+          responseAs[Option[Run]] shouldEqual runRespOption
+          responseAs[Option[Run]].size shouldEqual 1
```
myazinn: Redundant check. You already know that response is `Some(...)`, otherwise line 42 would have failed.

model/src/main/scala/cromwell/pipeline/model/wrapper/RunId.scala
```diff
@@ -21,4 +21,5 @@ object RunId extends Wrapped.Companion {
   )
 
   def random: RunId = new RunId(java.util.UUID.randomUUID().toString)
+  def convertRun(value: String): RunId = new RunId(value)
```
myazinn: The whole point of wrappers is that they won't allow someone to create object in illegal state, e.g. `RunId("definitely not UUID")`. This method bypasses these checks. You should remove it and use `RunId.from()` instead.

services/src/main/scala/cromwell/pipeline/service/RunService.scala
```diff
@@ -0,0 +1,50 @@
...
+class RunService(runRepository: RunRepository)(implicit executionContext: ExecutionContext) {
+
+  def addRun(runCreateRequest: RunCreateRequest): Future[RunId] = {
+
+    val newRun = Run(
+      runId = RunId.random,
+      projectId = runCreateRequest.projectId,
+      projectVersion = runCreateRequest.projectVersion,
+      status = Created,
+      timeStart = new Date().toInstant,
```
myazinn:
```suggestion
      timeStart = Instant.now(),
```

services/src/main/scala/cromwell/pipeline/service/RunService.scala
```diff
@@ -0,0 +1,50 @@
...
+      userId = runCreateRequest.userId,
+      results = runCreateRequest.results
+    )
+
+    runRepository.addRun(newRun)
+  }
+
+  def getRunById(runId: RunId): Future[Option[Run]] = runRepository.getRunById(runId)
```
myazinn: Shouldn't we forbid to access `Run` if current user is not an author of this `Run`?

myazinn (changes requested):
Great job!
I have a couple of small comments on this.

controllers/src/test/scala/cromwell/pipeline/controller/RunControllerTest.scala
manvicha: Line 43 removed

model/src/main/scala/cromwell/pipeline/model/wrapper/RunId.scala
manvicha: This is good to know. I was not sure how to get the `RunId` object that was the reason i created the method. Thanks for pointing me to use the correct syntax, i have applied `from` method

services/src/main/scala/cromwell/pipeline/service/RunService.scala
manvicha: Changed to `Instant.now()`

services/src/main/scala/cromwell/pipeline/service/RunService.scala
manvicha: I added the `userId` to the method

controllers/src/main/scala/cromwell/pipeline/controller/RunController.scala
```diff
@@ -0,0 +1,57 @@
...
+
+class RunController(runService: RunService)(implicit executionContext: ExecutionContext) {
+
+  val route: AccessTokenContent => Route = accessToken =>
+    path("runs") {
+      concat(
+        get {
+          parameter('run_id.as[String]) { runId =>
+            {
+              onComplete(runService.getRunByIdAndUser(RunId.from(runId), accessToken.userId)) {
```
myazinn: `runService.getRunByIdAndUser` must take `RunId` (as the name implies) as a parameter, not `ValidationResult[Wrapper]`.
You can leverage deserialization to akka by changing
`parameter('run_id.as[String]) { runId =>`
to
`parameter('run_id.as[RunId]) { runId =>`
In that case, you will have to provide an implicit Unmarshaller.
Alternatively, you can pattern match on the result of the `RunId.from` right here, or you must make a method in a `runService` with a more appropriate name.

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/entry/RunEntry.scala
```diff
@@ -37,9 +37,8 @@ trait RunEntry { this: Profile with UserEntry with ProjectEntry with CustomsWith
 
   val runs = TableQuery[RunTable]
 
-  def getRunById = Compiled { runId: Rep[RunId] =>
+  def getRunByIdAndUser(runId: RunId, userId: UserId) =
```
myazinn, line 33: You forgot to use `userId`.

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/entry/RunEntry.scala
manvicha: `userId` added

controllers/src/main/scala/cromwell/pipeline/controller/RunController.scala
manvicha: I created a Unmarshaller function. I am not sure if this needs to be placed there

myazinn: It should be placed somewhere nearby, for example `cromwell.pipeline.controller.utils.FromStringUnmarshallers`. Also, if you make it implicit (e.g. `implicit val stringToRunId = ???`) then `'run_id.as[RunId]` will work as expected.

manvicha: I placed the function in the path that you suggested. Also, the implicit identifier. I am really amazed of how scala can treat this, happy to learn things about this.Thanks!

myazinn (approved):
LGTM

mnesvit (approved):

PR #119, opened 2020-11-06
Fix flaky AuthService test

myazinn:
Fixes problems with AuthServiceTest that caused other tests to fail.

### DIFF
--- a/services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
@@ -19,7 +19,6 @@ import org.scalamock.scalatest.MockFactory
 import org.scalatest.concurrent.ScalaFutures.whenReady
 import org.scalatest.{ Matchers, WordSpec }
 import pdi.jwt.algorithms.JwtHmacAlgorithm
-import org.mockito.Matchers.any
 import pdi.jwt.{ Jwt, JwtAlgorithm, JwtClaim }
 import play.api.libs.json.Json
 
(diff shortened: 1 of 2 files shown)


services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
```diff
@@ -37,9 +36,6 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
 
   import authConfig._
 
-  private val userRepository: UserRepository = stub[UserRepository]
-  private val authUtils: AuthUtils = stub[AuthUtils]
-  private val authService: AuthService = new AuthService(userRepository, authUtils)
```
myazinn, line 42: Sharing mock between different tests leads to a lot of pain and bugs. I've moved them to `AuthServiceTestContext` below to ensure that each test has it's own mock.

services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
```diff
@@ -149,23 +145,24 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
     }
 
     "responseFromUser" should {
-      "return AuthResponse" in {
-        (authUtils.getAuthResponse _ when (*, *, *)).returns(Some(any[AuthResponse]))
-        authService.responseFromUser(dummyUser) shouldBe Some(any[AuthResponse])
+      "return whatever getAuthResponse returns" taggedAs Service in new AuthServiceTestContext {
+        private val dummyResponse = Some(AuthResponse("", "", 1))
+        (authUtils.getAuthResponse _ when (*, *, *)).returns(dummyResponse)
+        authService.responseFromUser(dummyUser) shouldBe dummyResponse
```
myazinn, line 151: `any[AuthResponse]` caused problems with other tests. It makes sense, but I have no idea, why this test did not fail by itself.
Mockito is great!

PR #120, opened 2020-11-06
Fix flaky AuthService test

myazinn:
Fixes problems with AuthServiceTest that caused other tests to fail.

### DIFF
--- a/services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
@@ -19,7 +19,6 @@ import org.scalamock.scalatest.MockFactory
 import org.scalatest.concurrent.ScalaFutures.whenReady
 import org.scalatest.{ Matchers, WordSpec }
 import pdi.jwt.algorithms.JwtHmacAlgorithm
-import org.mockito.Matchers.any
 import pdi.jwt.{ Jwt, JwtAlgorithm, JwtClaim }
 import play.api.libs.json.Json
 
@@ -37,9 +36,6 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
 
   import authConfig._
 
-  private val userRepository: UserRepository = stub[UserRepository]
-  private val authUtils: AuthUtils = stub[AuthUtils]
-  private val authService: AuthService = new AuthService(userRepository, authUtils)
   private val userId = UserId.random
   private val userPassword = "Password213"
   private val incorrectUserPassword = "Password2134"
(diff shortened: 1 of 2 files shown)


services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
```diff
@@ -37,9 +36,6 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
 
   import authConfig._
 
-  private val userRepository: UserRepository = stub[UserRepository]
-  private val authUtils: AuthUtils = stub[AuthUtils]
-  private val authService: AuthService = new AuthService(userRepository, authUtils)
```
myazinn, line 42: Sharing mock between different tests leads to a lot of pain and bugs. I've moved them to `AuthServiceTestContext` below to ensure that each test has it's own mock.

services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
```diff
@@ -149,23 +145,24 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
     }
 
     "responseFromUser" should {
-      "return AuthResponse" in {
-        (authUtils.getAuthResponse _ when (*, *, *)).returns(Some(any[AuthResponse]))
-        authService.responseFromUser(dummyUser) shouldBe Some(any[AuthResponse])
```
myazinn, line 154: `any[AuthResponse]` caused problems with other tests. It makes sense, but I have no idea why this test did not fail by itself.
Mockito is great!

services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
```diff
@@ -175,18 +172,24 @@ class AuthServiceTest extends WordSpec with Matchers with MockFactory {
...
-      "throw the exception if a user isn't active" in {
+      "throw the exception if a user isn't active" taggedAs Service in new AuthServiceTestContext {
         authService.userIsActive(inactiveUser) shouldBe
           Some(InactiveUserException(AuthService.inactiveUserMessage))
       }
     }
   }
 
-  class RefreshTokenContext(lifetime: Long) {
+  class AuthServiceTestContext {
```
alesavin, line 186: typically this type of fixture-objects is traits (https://www.scalatest.org/user_guide/sharing_fixtures, Instantiating fixture-context objects), but here I see no problem with class

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTest.scala
```diff
@@ -28,7 +27,8 @@ class ProjectFileServiceTest extends AsyncWordSpec with Matchers with MockitoSug
     "validateFile" should {
       "return valid message to valid file" taggedAs Service in {
         val request = ProjectFileContent(correctWdl)
-        when(womTool.validate(correctWdl)).thenReturn(Right(any[WomBundle]))
+        val dummyBundle = WomBundle(None, Map.empty, Map.empty, Set.empty)
```
alesavin, line 30: unrelated change?

GrigoriiBerezin (approved):
If it's really work like that, it's strange :) Thank you for fixing that

services/src/test/scala/cromwell/pipeline/service/AuthServiceTest.scala
myazinn: I prefer traits for test fixtures too. I have used class here only because the tests already have a context based on the class.

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTest.scala
myazinn: Nope. This test also has a problem with a misplaced matcher `any[Foo]` and might fail eventually.

alesavin (approved):
LGTM

iliaonishchenko (approved):
It looks good!

PR #74, opened 2020-05-27
Epmlstrcmw-157: delete and get endpoints for project file

IaroslavTavchenkov:

PR #125, opened 2020-12-08
Epmlstrcmw-157: delete and get endpoints for typed client

IaroslavTavchenkov:
This pull request is dedicated to rewriting PR #74, where Get and Delete endpoints for gitlab were implemented.  
You can find docs for gitlab **deleteFile** and **getFile** **API** by this [link](https://docs.gitlab.com/ee/api/repository_files.html#get-file-from-repository)  
  
Now there are several problems:  
\- we need to know which response we get after gitlab delete method  
\- then we should make a formatter with reads and writes methods in Project.scala  
\- and finally check types in all delete method occurrences  
  
p.s. If we still got a problem with deleting one file, maybe we should investigate multiple file API by this [link](https://docs.gitlab.com/ee/api/commits.html#create-a-commit-with-multiple-files-and-actions)


iliaonishchenko:
Out of date.

PR #126, opened 2020-12-08
feat: Add documentation to run cromwell workflows

manvicha:
Fill the README with some samples to run workflows
on the aws cromwell instance. A hello world example
was detailed with its requests and responses.


myazinn (changes requested):
Awesome, thank you!
I have one comment though. Please create separate files `hello.wdl` and `input.json`, instead of describing them in readme. I guess you can create a new folder in a root, similar to how it's done for Postman collections or scripts, for example.

mnesvit (approved):

myazinn (approved):

PR #128, opened 2020-12-19
Fix model for GitLab response

myazinn:
This PR solves two issues.
1. `id` of a project in GitLab is an `Integer` (see https://docs.gitlab.com/ee/api/projects.html#get-single-project), not a `String`. In Cromwell Pipeline we treated it as a `String`. This caused the app to say that the project was not created even though it was created successfully.
2. Some of the `ProjectService` methods said that they return `Future[Option[Project]]`, but there was no way to get `Future(None)` from them. Also, there was a lot of duplication when checking user rights to read / modify `Project`.

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
@@ -4,8 +4,8 @@ import java.util.UUID
 
 import cromwell.pipeline.datastorage.dao.repository.ProjectRepository
 import cromwell.pipeline.datastorage.dto._
-import cromwell.pipeline.service.Exceptions.{ ProjectAccessDeniedException, ProjectNotFoundException }
 import cromwell.pipeline.model.wrapper.UserId
+import cromwell.pipeline.service.Exceptions.{ ProjectAccessDeniedException, ProjectNotFoundException }
 
 import scala.concurrent.{ ExecutionContext, Future }
 
@@ -13,17 +13,18 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
   implicit executionContext: ExecutionContext
 ) {
 
-  def getProjectById(projectId: ProjectId): Future[Option[Project]] =
-    projectRepository.getProjectById(projectId)
+  private[service] def getUserProjectById(projectId: ProjectId, userId: UserId): Future[Project] =
+    projectRepository.getProjectById(projectId).flatMap(getForUserOrFail(_, userId))
 
-  def getProjectByName(namePattern: String, userId: UserId): Future[Option[Project]] = {
-    val result = projectRepository.getProjectByName(namePattern)
-    result.flatMap {
-      case Some(project) if project.ownerId == userId => result
+  def getUserProjectByName(namePattern: String, userId: UserId): Future[Project] =
+    projectRepository.getProjectByName(namePattern).flatMap(getForUserOrFail(_, userId))
+
+  private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
+    project match {
+      case Some(project) if project.ownerId == userId => Future.successful(project)
       case Some(_)                                    => Future.failed(new ProjectAccessDeniedException)
       case None                                       => Future.failed(new ProjectNotFoundException)
     }
-  }
 
   def addProject(request: ProjectAdditionRequest, userId: UserId): Future[Either[VersioningException, ProjectId]] = {
     val project =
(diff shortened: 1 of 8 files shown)


mnesvit (approved):

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -4,26 +4,32 @@ import java.util.UUID
...
+  def getProjectByName(namePattern: String, userId: UserId): Future[Project] =
+    projectRepository.getProjectByName(namePattern).map(getForUserOrFail(_, userId)).flatMap(Future.fromTry)
 
-  def getProjectByName(namePattern: String, userId: UserId): Future[Option[Project]] = {
-    val result = projectRepository.getProjectByName(namePattern)
-    result.flatMap {
-      case Some(project) if project.ownerId == userId => result
-      case Some(_)                                    => Future.failed(new ProjectAccessDeniedException)
-      case None                                       => Future.failed(new ProjectNotFoundException)
+  private def getForUserOrFail(project: Option[Project], userId: UserId): Try[Project] =
```
alesavin: why here we switched to Try? If you continue with Future[Project] it will be less transformations (Future.fromTry), so like

```
 project match {
      case Some(project) if project.ownerId == userId => Future.successful(project)
      ...
```

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -39,32 +45,25 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
...
-      case Some(_) =>
-        for {
-          _ <- projectRepository.deactivateProjectById(projectId)
-          getProject <- getProjectById(projectId)
-        } yield getProject
-      case None => Future.failed(new ProjectNotFoundException)
+  def deactivateProjectById(projectId: ProjectId, userId: UserId): Future[Project] =
+    getProjectById(projectId, userId).flatMap { project =>
+      if (!project.active) {
+        Future.failed(ProjectIllegalStateException(s"Cannot deactivate non-active project"))
```
alesavin: why we should fail if project already deactivated? just do nothing and return project?

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -4,26 +4,32 @@ import java.util.UUID
...
 import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success, Try }
 
 class ProjectService(projectRepository: ProjectRepository, projectVersioning: ProjectVersioning[VersioningException])(
   implicit executionContext: ExecutionContext
 ) {
 
-  def getProjectById(projectId: ProjectId): Future[Option[Project]] =
-    projectRepository.getProjectById(projectId)
+  private[service] def getProjectById(projectId: ProjectId, userId: UserId): Future[Project] =
```
alesavin: I suggest to rename this to getUserProjectById

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -4,26 +4,32 @@ import java.util.UUID
...
+    projectRepository.getProjectById(projectId).map(getForUserOrFail(_, userId)).flatMap(Future.fromTry)
+
+  def getProjectByName(namePattern: String, userId: UserId): Future[Project] =
```
alesavin: getUserProjectByName?

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: It's just my personal preference. If we have `Try`, we can easily reuse it and integrate into a code that uses `Future` or `Option`. If we have `Future`, it's not that easy to integrate it with `Try` or `Option`.
But since it is a private method and these transformations are ugly, I think Future[Project] should be fine too.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Done

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: Done

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
myazinn: IDK, I thought deactivation of a deactivated project should not be allowed. Anyway, I decided to return as it was and not return an error in this case.

alesavin (approved):

PR #129, opened 2020-12-27
EPMLSTRCMW-206 Explicitly store and use GitLab ID

myazinn:
Explicitly store GitLab ID as `repositoryId`
Forbid to change `repositoryId`
Add RunController routes to all routes
Make `repositoryId` required field in `Project`

### DIFF
--- a/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
+++ b/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
@@ -86,4 +86,15 @@
         <addNotNullConstraint constraintName="results_nn" tableName="run" columnName="results"/>
     </changeSet>
 
+    <changeSet id="change_repository_id_type" author="nikita_miazin">
+        <delete tableName="run"/>
+        <delete tableName="project"/>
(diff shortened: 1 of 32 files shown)


datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -86,4 +86,15 @@
         <addNotNullConstraint constraintName="results_nn" tableName="run" columnName="results"/>
     </changeSet>
 
+    <changeSet id="change_repository_id_type" author="nikita_miazin">
+        <delete tableName="run"/>
+        <delete tableName="project"/>
```
myazinn, line 91: These two lines will delete all projects in Cromwell Pipeline. I don't think this is a big deal, since we can easily create a new one.
It's not that hard to make a migration that will parse existing `repository` field, but I really don't want to do it.

mnesvit (approved):

GrigoriiBerezin (approved):

Dmitrii-Shevchenko (approved):
Great work!

PR #132, opened 2021-02-16
fix: EPMLSTRCMW-180 User can create invalid password

eshenko:


myazinn (commented):
Regex changes look good to me, but why do you need to change HTTP method from `put` to `post`?

eshenko:
> Regex changes look good to me, but why do you need to change HTTP method from `put` to `post`?

There were two equal Put endpoints in the UserController. For updating user information and for changing user's password. When we send an incorrect password, we go into the updating user information endpoint, instead of going to the changing user's password endpoint.

I had two options:
1. To add some path (for example, users/info and users/password)
2. To change the method of one of the endpoints

I chose the second one. Am I wrong?

eshenko:
Fixed.
Added paths `users/info` and `users/password` in the Put methods of the UserController.

controllers/src/main/scala/cromwell/pipeline/controller/UserController.scala
```diff
@@ -14,32 +14,38 @@ import scala.util.{ Failure, Success }
...
-    path("users") {
-      concat(
-        get {
-          parameter('email.as[String]) { email =>
-            onComplete(userService.getUsersByEmail(email)) {
-              case Success(r) => complete(r)
-              case Failure(exc) =>
-                complete(StatusCodes.InternalServerError, exc.getMessage)
+    concat(
+      path("users") {
```
myazinn: Let's make it DRY. You don't need concatenation here and can reuse `path("users")` for `put` endpoints
```
path(users) {
  concat (
    get { },
    delete { },
    path(info) { put { } },
    path(password) { put { } }
  )
}
```
or
```
path(users) {
  concat (
    get { },
    delete { },
    put {
      concat(
        path(info) { },
        path(password) { }  
      )
    }
  )
}
```
Which one to choose is up to you.

eshenko: fixed.

GrigoriiBerezin (approved):
Nice

myazinn (approved):

PR #135, opened 2021-02-25
[Suspended] feat: EPMLSTRCMW-182 Separately filter unsuccessful auth

eshenko:
In order to get log messages in `userEmail : message` format I had to move object validation calls into services.

Earlier validations were called automatically while unmarshalling was running, it made getting log messages in the correct format impossible.

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/AuthService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/AuthService.scala
@@ -1,8 +1,9 @@
 package cromwell.pipeline.service
 
-import java.time.Instant
+import cats.Show
 
-import cats.data.OptionT
+import java.time.Instant
+import cats.data.{ NonEmptyChain, Validated }
 import cats.implicits._
 import cromwell.pipeline.auth.AuthUtils
 import cromwell.pipeline.datastorage.dao.repository.UserRepository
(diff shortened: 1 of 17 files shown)


model/src/main/scala/cromwell/pipeline/model/validator/Wrapped.scala
```diff
@@ -4,40 +4,34 @@ import cats.Show
...
+import play.api.libs.json.{ Format, JsResult, JsSuccess, JsValue, Json }
 
 import scala.util.control.NoStackTrace
 import scala.language.implicitConversions
+import org.slf4j.LoggerFactory
 
-trait Wrapped[T] extends Any {
-  def unwrap: T
+trait Wrapped extends Any {
+  def unwrap: String
```
myazinn: I haven't watched the whole request yet, but I accidentally noticed it. I don't know what you are trying to do, but you are doing it wrong.
`Wrapped` should be able to wrap values of any given type and provide some useful logic.

eshenko: fixed.

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -15,39 +15,74 @@ import cromwell.pipeline.datastorage.dto.auth.{
...
-      case Some(_) => Future.failed(DuplicateUserException(s"${request.email} already exists"))
+  def signUp(request: SignUpRequest): Future[Option[AuthResponse]] = {
+    val data: Either[NonEmptyChain[String], Future[Option[AuthResponse]]] = for {
+      email <- logged(
+        UserEmail.from(request.email),
+        "Registration failure with email " + request.email + " : invalid email"
+      ).toEither
+      firstname <- logged(
+        Name.from(request.firstName),
+        "Registration failure with email " + email + " : invalid firstname"
```
myazinn: 1) Prefer string interpolation over string concatenation. String concatenation has some drawbacks that might bite you
2) Probably it would be useful to include invalid values to log messages (except for password)
3) Create a helper method to avoid duplication 
```
def failMsg(cause: String): String = s""Registration failure with email ${request.email} : $cause"

failMsg("invalid first name [${request.firstname}]")
```

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -87,24 +133,29 @@ class AuthService(userRepository: UserRepository, authUtils: AuthUtils)(implicit
...
+    validated.leftMap { e =>
+      log.warn(s"$error. Reason: [${e.show}]")
+      e
+    }
 }
 
 object AuthService {
-  val authorizationFailure = "invalid email or password"
-  val inactiveUserMessage = "User is not active"
+  val log: Logger = LoggerFactory.getLogger(AuthService.getClass)
```
myazinn: It's a good practice to create a logger inside a class, not the companion object.
In general, you should avoid usage of objects that perform IO if they can't be overwritten. It makes them hard to test and impossible to re-use.

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
```diff
@@ -87,17 +138,23 @@ class AuthService(userRepository: UserRepository, authUtils: AuthUtils)(implicit
...
-    if (user.passwordHash == StringUtils.calculatePasswordHash(request.password.value, user.passwordSalt)) None
-    else Some(IncorrectPasswordException(authorizationFailure))
+  def passwordCorrect(password: Password, user: User): Option[Throwable] =
+    if (user.passwordHash == StringUtils.calculatePasswordHash(password.value, user.passwordSalt)) None
+    else Some(IncorrectPasswordException("incorrect password"))
 
   def userIsActive(user: User): Option[Throwable] =
-    if (user.active) None else Some(InactiveUserException(inactiveUserMessage))
+    if (user.active) None
+    else Some(InactiveUserException("user is not active"))
```
myazinn: Named strings (like `AuthService.inactiveUserMessage`) are much better than raw strings somewhere in the middle of the code.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -23,30 +28,86 @@ class UserService(userRepository: UserRepository)(implicit executionContext: Exe
...
+            "Password updating failure with email " + user.email + " : invalid current password"
+          ).toEither
+          newPassword <- logged(
+            Password.from(request.newPassword),
+            "Password updating failure with email " + user.email + " : invalid new password"
+          ).toEither
+          repeatPassword <- logged(
+            Password.from(request.repeatPassword),
+            "Password updating failure with email " + user.email + " : invalid repeat password"
+          ).toEither
```
myazinn: Validate that passwords are valid before reading a user from DB. Validation is fast unlike DB operations.

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
eshenko: Good advice, thanks! Fixed.

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
eshenko: Fixed.

services/src/main/scala/cromwell/pipeline/service/AuthService.scala
eshenko: Fixed.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
eshenko: I chose this order because I use a user email that gets from DB in log messages.
Should I change it?

myazinn:
I just remembered we decided put this MR on hold until we find a better way to implement it. Right now it looks too cumbersome and we don't know how to improve it so let's suspend it for a while.

myazinn:
Since we couldn't find a neater way to do it in reasonable time, I'm closing this PR.
We can reopen it later, but I don't think we'll need it shortly.

PR #136, opened 2021-03-01
feat: EPMLSTRCMW-182 Wrapped generics instead abstract

eshenko:
For resolving EPMLSTRCMW-182 We need to use generic types in the Wrapped.Companion trait instead of abstract types.


myazinn (approved):

eshenko:
out of date now

PR #141, opened 2021-03-23
fix: EPMLSTRCMW-216-Create-GitLabFileContent

eshenko:


services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -225,14 +225,15 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
+      .get[GitLabFileContent](
+        s"${config.url}/projects/${project.repositoryId.value}/repository/files/$filePath",
         Map("ref" -> fileVersion),
         config.token
       )
       .map {
-        case Response(HttpStatusCodes.OK, SuccessResponseBody(projFileContent), _) =>
-          Right(ProjectFile(path, projFileContent))
+        case Response(HttpStatusCodes.OK, SuccessResponseBody(gitLabFile), _) =>
+          val content = new String(java.util.Base64.getDecoder.decode(gitLabFile.content))
```
myazinn: I think you should create a method like
`private def decodeBase64(str: String): String = ???`
It will make code a little bit easier to read and will allow us to override this method in tests if necessary

eshenko: Method created.

iliaonishchenko (approved):
Looks good to me 👍

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -166,6 +166,14 @@ object ProjectFileContent {
   implicit val projectFileContentFormat: Format[ProjectFileContent] = Json.valueFormat[ProjectFileContent]
 }
 
+final case class GitLabFileContent(content: String)
+
+object GitLabFileContent {
+  implicit val encodingProjectFileContentFormat: Reads[GitLabFileContent] = {
+    (JsPath \ "content").read[String].map(GitLabFileContent.apply)
```
GrigoriiBerezin:
```suggestion
  implicit lazy val projectFileContentFormat: OFormat[GitLabFileContent] = Json.format[GitLabFileContent]
```
should work as well I think

GrigoriiBerezin:
Nice one!

GrigoriiBerezin (approved):
Nice one!

PR #144, opened 2021-03-30
fix: change logic of adding project configuration

eshenko:


repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
```diff
@@ -6,13 +6,19 @@ import ProjectFile.pathFormat
 import org.mongodb.scala.bson.collection.immutable.Document
 import play.api.libs.json.{ Json, OFormat }
 
-case class ProjectFileConfiguration(path: Path, inputs: List[FileParameter])
+case class ProjectFileConfiguration(path: Path, inputs: Set[FileParameter]) {
+   override def equals(obj: Any): Boolean = obj match {
+     case o: ProjectFileConfiguration => this.path == o.path
+     case _ => false
+   }
+   override def hashCode(): Int = path.hashCode()
```
myazinn: I disagree with this definition.

If we have two configurations for the same file, it doesn't mean the are the same. The whole point of ProjectFile**Configuration** is to have the configurations. It doesn't make sense to not use them in `equals` / `hashcode`implementations.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -8,10 +8,22 @@ import scala.concurrent.{ ExecutionContext, Future }
...
+      repository
+        .updateOne(toDocument(configuration), "projectId", projectConfiguration.projectId.value)
+        .map(_.toString)
+    }
+
+    getById(projectConfiguration.projectId) flatMap {
+      case Some(config) if projectConfiguration.projectFileConfigurations != config.projectFileConfigurations =>
+        val oldConfig = config.projectFileConfigurations diff projectConfiguration.projectFileConfigurations
+        val newConfig = projectConfiguration
+          .copy(projectFileConfigurations = oldConfig ++ projectConfiguration.projectFileConfigurations)
```
myazinn:
```suggestion
       val newFileConfigurations = projectConfiguration.projectFileConfigurations ++ config.projectFileConfigurations)
        val newConfig = projectConfiguration.copy(projectFileConfigurations = newFileConfigurations)
```
Although this line most likely will be changed because of problems with `equals` / `hashcode` definitions

services/src/test/scala/cromwell/pipeline/service/ProjectConfigurationServiceTest.scala
```diff
@@ -17,23 +17,55 @@ class ProjectConfigurationServiceTest extends AsyncWordSpec with Matchers with M
...
+        when(configurationRepository.getByParam("projectId", projectId))
+          .thenReturn(Future.successful(List(activeDocument)))
+        when(updateResult.toString).thenReturn("Success update")
+        when(configurationRepository.updateOne(newActiveDocument, "projectId", projectId))
+          .thenReturn(Future.successful(updateResult))
+
+        configurationService.addConfiguration(newConfiguration).map(_ shouldBe "Success update")
+      }
+
+      "return complete status for adding new configuration" in {
```
myazinn: I wonder why you needed to change an existing test.
If you did everything correctly, then you shouldn't change existing test for configuration creation. Why is it changed?

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
eshenko: Removed this overriding

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
eshenko: Changed realization.

services/src/test/scala/cromwell/pipeline/service/ProjectConfigurationServiceTest.scala
eshenko: That's because the old test just replaces the configuration without checking if configuration with the same path exists.

myazinn: I get it, but it should not affect test implementation that much. You should just add one more mocked call for `getByParam` method and it should be enough.

eshenko: Fixed.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
```diff
@@ -6,16 +6,24 @@ import ProjectFile.pathFormat
...
 
 object ProjectFileConfiguration {
   implicit val projectFileConfigurationFormat: OFormat[ProjectFileConfiguration] = Json.format
+
+  def toMap(config: ProjectFileConfiguration): Map[Path, Map[String, TypedValue]] =
+    Map(
+      config.path -> config.inputs.flatMap(fileParameter => Map(fileParameter.name -> fileParameter.typedValue)).toMap
+    )
+
+  def fromMap(path: Path, inputs: Map[String, TypedValue]): ProjectFileConfiguration =
```
myazinn: `fromMap` name looks misleading. We are creating `FileParameter` from a map, but `ProjectFileConfiguration` is created using two regular parameters.
If we have methods like `toSmth` / `fromSmth` then I would say the following code should work
```
val config1: ProjectFileConfiguration = ???
val config2: ProjectFileConfiguration = fromMap(toMap(config1))
```
But your implementation won't compile which is extremely confusing.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
```diff
@@ -6,16 +6,24 @@ import ProjectFile.pathFormat
...
+    ProjectFileConfiguration(path, inputs.flatMap(x => Set(FileParameter(x._1, x._2))).toSet)
 }
 
 case class ProjectConfiguration(
   projectId: ProjectId,
   active: Boolean,
-  projectFileConfigurations: List[ProjectFileConfiguration]
+  projectFileConfigurations: Set[ProjectFileConfiguration]
```
myazinn: `Set` here is rather useless. It won't save us from cases when we have two configurations with equal paths but different inputs.
If you want to make sure there is no duplicate configurations then let's use `Map` directly.
Something like 
```suggestion
  configurations: Map[Path, Configuration]
```
where Configuration is
```
case class Configuration(inputs: Set[FileParameter])
```
should be good enough.
Or we can leave it as is. I just don't think using `Set` here is a good idea because it gives a false perception that these configurations would be unique the way we want them to be.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
```diff
@@ -6,16 +6,24 @@ import ProjectFile.pathFormat
...
-case class ProjectFileConfiguration(path: Path, inputs: List[FileParameter])
+case class ProjectFileConfiguration(path: Path, inputs: Set[FileParameter])
 
 object ProjectFileConfiguration {
   implicit val projectFileConfigurationFormat: OFormat[ProjectFileConfiguration] = Json.format
+
+  def toMap(config: ProjectFileConfiguration): Map[Path, Map[String, TypedValue]] =
+    Map(
+      config.path -> config.inputs.flatMap(fileParameter => Map(fileParameter.name -> fileParameter.typedValue)).toMap
+    )
```
myazinn: I don't think this is an appropriate place for such method (same for `fromMap`). It solves a very specific problem and it's better to put it where you actually need it.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,33 @@
...
+    getById(projectConfiguration.projectId).flatMap {
+      case Some(config) if projectConfiguration.projectFileConfigurations != config.projectFileConfigurations =>
+        val oldFileConfigsMap =
+          config.projectFileConfigurations.flatMap(ProjectFileConfiguration.toMap).toMap
+        val newFileConfigsMap =
+          projectConfiguration.projectFileConfigurations.flatMap(ProjectFileConfiguration.toMap).toMap
+
+        val updatedFileConfigsMap = oldFileConfigsMap ++ newFileConfigsMap
+        val updatedFileConfigsSet =
+          updatedFileConfigsMap.flatMap(x => Set(ProjectFileConfiguration.fromMap(x._1, x._2))).toSet
```
myazinn: Avoid meaningless variables especially when you work with tuples. It's hard to understand what `x._2` means.
```suggestion
          updatedFileConfigsMap.flatMap { case (path, inputs) => Set(ProjectFileConfiguration.fromMap(path, inputs))).toSet
```

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
eshenko: Renamed and moved these methods from here.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
eshenko: Very useful, thanks!
Fixed.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
eshenko: Moved.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectConfiguration.scala
eshenko: I faced problems with marshaling/unmarshaling when I used `configurations: Map[Path, Configuration]`. We need to use List, not Map.
So I decided to leave it as is.

myazinn: If you faced some problems then why didn't you ask how to solve them? :)
Take a look at `play.api.libs.json.DefaultWrites#keyMapWrites` and `play.api.libs.json.DefaultReads#mapReads`

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,49 @@
...
+
+    def toMap(config: ProjectFileConfiguration): Map[Path, Map[String, TypedValue]] =
+      Map(
+        config.path -> config.inputs.flatMap(fileParameter => Map(fileParameter.name -> fileParameter.typedValue)).toMap
+      )
+
+    def toProjectFileConfiguration(path: Path, inputs: Map[String, TypedValue]): ProjectFileConfiguration =
+      ProjectFileConfiguration(path, inputs.flatMap {
+        case (name, typedValue) => List(FileParameter(name, typedValue))
+      }.toList)
```
myazinn:
```suggestion
      ProjectFileConfiguration(path, inputs.map((FileParameter.apply _).tupled).toList)
```
Now once it's clear what you want to do, we can simplify it

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,49 @@
...
+
+    getById(projectConfiguration.projectId).flatMap {
+      case Some(config) if projectConfiguration.projectFileConfigurations != config.projectFileConfigurations =>
+        val oldFileConfigsMap = config.projectFileConfigurations.flatMap(toMap).toMap
+        val newFileConfigsMap = projectConfiguration.projectFileConfigurations.flatMap(toMap).toMap
+
+        val updatedFileConfigsMap = oldFileConfigsMap ++ newFileConfigsMap
+        val updatedFileConfigs = updatedFileConfigsMap.flatMap {
+          case (path, inputs) => List(toProjectFileConfiguration(path, inputs))
+        }.toList
```
myazinn:
```suggestion
        val updatedFileConfigs = updatedFileConfigsMap.map((toProjectFileConfiguration _).tupled).toList
```

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,49 @@
...
-    repository
-      .updateOne(toDocument(projectConfiguration), "projectId", projectConfiguration.projectId.value)
-      .map(_.toString)
+  def addConfiguration(projectConfiguration: ProjectConfiguration): Future[String] = {
+    def updateConfiguration(configuration: ProjectConfiguration) =
+      repository.updateOne(toDocument(configuration), "projectId", projectConfiguration.projectId.value).map(_.toString)
+
+    def toMap(config: ProjectFileConfiguration): Map[Path, Map[String, TypedValue]] =
+      Map(
+        config.path -> config.inputs.flatMap(fileParameter => Map(fileParameter.name -> fileParameter.typedValue)).toMap
```
myazinn:
```suggestion
        config.path -> config.inputs.map(fileParameter => fileParameter.name -> fileParameter.typedValue).toMap
```

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,49 @@
...
 
-  def addConfiguration(projectConfiguration: ProjectConfiguration): Future[String] =
-    repository
-      .updateOne(toDocument(projectConfiguration), "projectId", projectConfiguration.projectId.value)
-      .map(_.toString)
+  def addConfiguration(projectConfiguration: ProjectConfiguration): Future[String] = {
+    def updateConfiguration(configuration: ProjectConfiguration) =
+      repository.updateOne(toDocument(configuration), "projectId", projectConfiguration.projectId.value).map(_.toString)
+
+    def toMap(config: ProjectFileConfiguration): Map[Path, Map[String, TypedValue]] =
```
myazinn: I thought you will use a Map as a main data structure and get rid of this method but if you want to keep it then let's make it more readable.
Can't we just return `(Path, Map[String, TypedValue])` ? Usually we should not return tuples because they make code unreadable but in this case `Map[Path, Map[String, TypedValue]]` is no better than `(Path, Map[String, TypedValue])`

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,49 @@
...
+        val newFileConfigsMap = projectConfiguration.projectFileConfigurations.flatMap(toMap).toMap
+
+        val updatedFileConfigsMap = oldFileConfigsMap ++ newFileConfigsMap
+        val updatedFileConfigs = updatedFileConfigsMap.flatMap {
+          case (path, inputs) => List(toProjectFileConfiguration(path, inputs))
+        }.toList
+
+        val newConfig = projectConfiguration.copy(projectFileConfigurations = updatedFileConfigs)
+        updateConfiguration(newConfig)
+      case _ => updateConfiguration(projectConfiguration)
```
myazinn: Sorry for not bringing it up earlier but are we sure we want to update a configuration if we already know that old and new configurations are identical? I mean this default case can be achieved either if there is no configuration or if there is a configuration but it's the same as we want to put. Maybe we should not do anything in this case? Alternatively you case remove guardian condition in the first `case`

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,45 @@
...
+
+    def toTuple(config: ProjectFileConfiguration): (Path, Map[String, TypedValue]) =
+      config.path -> config.inputs.map(fileParameter => fileParameter.name -> fileParameter.typedValue).toMap
+
+    def toProjectFileConfiguration(path: Path, inputs: Map[String, TypedValue]): ProjectFileConfiguration =
+      ProjectFileConfiguration(path, inputs.map((FileParameter.apply _).tupled).toList)
+
+    getById(projectConfiguration.projectId).flatMap {
+      case Some(config) =>
+        val oldFileConfigsMap = config.projectFileConfigurations.flatMap(fileConfig => Map(toTuple(fileConfig))).toMap
```
myazinn:
```suggestion
        val oldFileConfigsMap = config.projectFileConfigurations.map(toTuple).toMap
```

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,17 +1,45 @@
...
+        val newFileConfigsMap =
+          projectConfiguration.projectFileConfigurations.flatMap(fileConfig => Map(toTuple(fileConfig))).toMap
```
myazinn:
```suggestion
          projectConfiguration.projectFileConfigurations.map(toTuple).toMap
```

myazinn (approved):
Looks good to me, but please stop doing `flatMap` when `map` is more than enough.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
eshenko: Thanks! Fixed.

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
eshenko: Thanks! Fixed.

Dmitrii-Shevchenko (approved):

PR #147, opened 2021-04-05
EPMLSTCMW-221 migrate from travis to GitHub actions

myazinn:


project/plugins.sbt
```diff
@@ -2,5 +2,6 @@ addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
 addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.8.0")
 addSbtPlugin("com.github.cb372" % "sbt-explicit-dependencies" % "0.2.16")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.1")
-addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.2.7")
+//addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.2.7")
```
Dmitrii-Shevchenko, line 5: We still need it?

myazinn: I would like to leave it here for a while.
Most likely we will have to remove `coveralls` completely because it doesn't work well with `sbt`. But for now I want to explicitly show that we are aware of official plugin but it doesn't suite us.

Dmitrii-Shevchenko (approved):

controllers/src/main/scala/cromwell/pipeline/controller/ProjectController.scala
```diff
@@ -19,13 +19,14 @@ class ProjectController(projectService: ProjectService)(
...
-              case Failure(e)                               => complete(StatusCodes.InternalServerError, e.getMessage)
-            }
+          parameter('name.as[String]) {
+            name =>
+              onComplete(projectService.getUserProjectByName(name, accessToken.userId)) {
+                case Success(project)                         => complete(project)
+                case Failure(e: ProjectNotFoundException)     => complete(StatusCodes.NotFound, e.getMessage)
+                case Failure(e: ProjectAccessDeniedException) => complete(StatusCodes.Forbidden, e.getMessage)
+                case Failure(e)                               => complete(StatusCodes.InternalServerError, e.getMessage)
+              }
```
GrigoriiBerezin, line 29: Scalafmt config is different in team again, I suppose :)

GrigoriiBerezin (approved):
Nice PR, awesome code

controllers/src/main/scala/cromwell/pipeline/controller/ProjectController.scala
myazinn: Build was failing until I changed it.
I have no idea why it worked fine before, because I get the same changes when I run `sbt fmt` on the develop branch.

.github/workflows/cmwl_pipeline_ci.yml
```diff
@@ -0,0 +1,49 @@
...
+    branches:
+      - develop
+      - master
+  pull_request:
+    branches:
+      - develop
+      - master
+
+jobs:
+  lint:
```
iliaonishchenko: Why do we need two jobs?
Why don't we do lint check inside build?

I'm asking, because we're doing Java and Scala setup in both jobs, each setup takes ~10 seconds. 
If we do lint check in the build job, we will save these seconds...

myazinn: I wanted to make them separate jobs because it's a quite common mistake to not format code appropriately. It would have been nice to quickly see that your code is badly formatted instead of going to build details to see what went wrong.
Anyway, now it's changed and build is faster than ever.

iliaonishchenko (approved):
Wow that's great and I LOVE this MR!

PR #148, opened 2021-04-06
feat: run Pipeline using Docker

eshenko:


portal/docker.sbt
```diff
@@ -0,0 +1,21 @@
+defaultLinuxInstallLocation in Docker := "/pipeline/app"
+packageName in Docker := "pipeline"
```
myazinn: Let's make a bit more unique name, e.g. `cmwl_pipeline`

portal/src/main/resources/application.conf
```diff
@@ -1,5 +1,6 @@
 webservice {
   interface = "localhost"
+  interface = ${?HOSTNAME}
```
myazinn, line 5: Do we need it? I don't see how you use it. If we need it, consider renaming it to follow conventions that we use for other env variables (like `CMWL_DB_SERVVER_NAME`)

myazinn (changes requested):

Dmitrii-Shevchenko (approved):
Great work!

portal/src/main/resources/application.conf
Dmitrii-Shevchenko: We need this because docker has no "localhost" inside and it generates its own host (like http://e30590c064c6:8080/)

myazinn: Got it. Okay then, let's just give it a better name.

eshenko: It's not my own variable, it's the default variable in the docker image which our image based on. So I can't change its name

portal/docker.sbt
eshenko: Fixed.

portal/src/main/resources/application.conf
Dmitrii-Shevchenko: I think you can replace second "interface" with "docker_interface" and add check for this variable here 
[**https://github.com/EpamLifeSciencesTeam/cmwl_pipeline/blob/develop/utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala#L40**](url)
@myazinn What do you think?

GrigoriiBerezin (approved):
Nice

portal/src/main/resources/application.conf
myazinn: @Dmitrii-Shevchenko
I don't think it will make much of a difference . In fact, if we have to change application code for different deployment methods then it's even worse.
@eshenko
Leave it as is but add a comment on why we use this particular variable.

eshenko: Done.

myazinn (approved):
We are starting to look like a serious project!

PR #150, opened 2021-04-13
fix: Update regex for commit message git hook

egor-ka:


myazinn:
Outdated. We decided to remove local git hooks completely.

PR #151, opened 2021-04-20
EPMLSTRCMW-220 Running the pipeline via docker compose

Dmitrii-Shevchenko:
[https://jirapct.epam.com/jira/browse/EPMLSTRCMW-220](EPMLSTRCMW-220)
Moved some environment variables from docker.sbt to .env
Created .env.example, which should be an example for the .env file that docker loads for containers
Slightly changed readme description about running containers

Please check how it works on Mac/Linux, especially with Liquibase migrations


README.md
```diff
@@ -123,9 +123,9 @@ Run the next commands
 
 ## Run Cromwell Pipeline in the Docker (Local)
 
-- Change values of the environment variables in [docker.sbt](portal/docker.sbt) file
+- Create .env file with environment variables from .env.example file in the root folder
```
myazinn, line 126: [minor] Probably we should add this `.env` file to `.gitignore`

eshenko (approved):
Nice improvement 👍

GrigoriiBerezin (approved):
Niceee

README.md
Dmitrii-Shevchenko: You are right! Did it

portal/docker.sbt
```diff
@@ -1,21 +1,3 @@
-defaultLinuxInstallLocation in Docker := "/cmwl_pipeline/app"
-packageName in Docker := "cmwl_pipeline"
-dockerExposedVolumes := Seq("/cmwl_pipeline/app/logs")
-dockerExposedPorts := Seq(8080)
```
myazinn, line 4: When I run `sbt clean docker:publishLocal` it warns me
```
[warn] [1] There are no exposed ports for your docker image
[warn]  Configure the `dockerExposedPorts` or `dockerExposedUdpPorts` setting. E.g.
[warn] 
[warn]  // standard tcp ports
[warn]  dockerExposedPorts ++= Seq(9000, 9001)
[warn] 
[warn]  // for udp ports
[warn]  dockerExposedUdpPorts += 4444
```
Shouldn't we keep ports settings here?

.env.example
```diff
@@ -0,0 +1,18 @@
+CMWL_PIPELINE_CONTAINER_PATH=/cmwl_pipeline/app
+CMWL_PIPELINE_CONTAINER_NAME=cmwl_pipeline
+CMWL_PIPELINE_CONTAINER_LOGS=/cmwl_pipeline/app/logs
+CMWL_DB_SERVER_NAME=172.17.0.1
+CMWL_DB_PORT_NUMBER=5432
```
myazinn, line 6: Apparently you should reuse this port in description of docker container for Postgres. I changed it to `5433` and application failed to start. Changing
```
    ports:
      - "5432:5432"
```
to
```
    ports:
      - ${CMWL_DB_PORT_NUMBER}:5432
```
in `docker-compose.yml` resolved this issue. I assume you should do the same thing for other variables like mongo DB port.

myazinn (changes requested):

portal/docker.sbt
Dmitrii-Shevchenko: Added exposed ports. I'm not sure about safety explicit cast to Int, but I think it's clear that the port should be Int in .env file. Do I need to add type checking?

.env.example
Dmitrii-Shevchenko: Done

portal/docker.sbt
myazinn: I don't think we have a way to recover from such cases so it should be ok.

myazinn:
Changes look to me, I'll check how they work on my laptop and then approve.

myazinn (approved):

PR #152, opened 2021-04-23
EPMLSTRCMW-214

OlegKoshelev:
Project versioning has been implemented.
- "version" column has been added to project table in DB
- version field has been added to Project and LocalProject classes
- Tests have been fixed
- changelog-0.2.xml script has been added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .idea
 **/target
 **/logs
+null/
--- a/datasource/src/main/resources/liquibase/changelog/changelog-master.xml
+++ b/datasource/src/main/resources/liquibase/changelog/changelog-master.xml
@@ -4,6 +4,6 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
-    <include file="changelog-0.1.xml" relativeToChangelogFile="true"/>
+    <include file="changelog-0.2.xml" relativeToChangelogFile="true"/>
 
 </databaseChangeLog>
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
@@ -3,6 +3,7 @@ package cromwell.pipeline.datastorage.dao.entry
 import cromwell.pipeline.datastorage.Profile
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
+import slick.lifted.MappedToBase.mappedToIsomorphism
 import slick.lifted.{ ForeignKeyQuery, ProvenShape }
 
 trait ProjectEntry { this: Profile with UserEntry with CustomsWithEnumSupport with AliasesSupport =>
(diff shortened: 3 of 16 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@
 .idea
 **/target
 **/logs
+null/
```
GrigoriiBerezin, line 7: Why do we need this?

datasource/src/main/resources/liquibase/changelog/changelog-master.xml
```diff
@@ -4,6 +4,6 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
                                        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.4.xsd">
 
-    <include file="changelog-0.1.xml" relativeToChangelogFile="true"/>
+    <include file="changelog-0.2.xml" relativeToChangelogFile="true"/>
```
GrigoriiBerezin, line 7: I'm not sure, but I think we should just add another changelog file and include it after first one. @myazinn what do you think?

repositories/src/main/scala/cromwell/pipeline/datastorage/dao/entry/ProjectEntry.scala
```diff
@@ -40,4 +43,10 @@ trait ProjectEntry { this: Profile with UserEntry with CustomsWithEnumSupport wi
 
   def updateProjectNameAction(updatedProject: Project): ActionResult[Int] =
     projects.filter(_.projectId === updatedProject.projectId).map(project => project.name).update(updatedProject.name)
+
+  def updateProjectVersionAction(updatedProject: Project): ActionResult[Int] =
+    projects
+      .filter(_.projectId === updatedProject.projectId)
+      .map(project => project.version)
```
GrigoriiBerezin, line 50: opt: could be changed to `.map(_.version)`

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -91,8 +94,15 @@ object GitLabVersion {
...
   import VersionValue._
+  import PipelineVersion._
   def name: String = s"v$major.$minor.$revision"
 
+//  def this(stringVersion: String) = this(
+//    PipelineVersion(stringVersion).major,
+//    PipelineVersion(stringVersion).minor,
+//    PipelineVersion(stringVersion).revision
+//  )
+
```
GrigoriiBerezin, line 105: Lines like this must be removed in final commit

GrigoriiBerezin (changes requested):
Nice work, but there are some assumptions for you. Also, you should cover `ProjectEntry.updateProjectVersionAction`, `ProjectRepository.updateProjectVersion`, `ProjectService.updateProjectVersion` with tests

datasource/src/main/resources/liquibase/changelog/changelog-master.xml
myazinn: We should definetely add another changelog or add new changeset to the existing one.
@OlegKoshelev Once migrations are applied, they should never be removed / replaced / changed. If you want to change DB, you should create a new migration that should work on top of existing migrations, not instead of them.

DruzhininPavel (approved):
Not everything is clear for me in this code.
But I am ok with this code.

Nikolai-Isaev:
Can't wrap my brains around a lot of it for now, seems good to me.

eshenko (changes requested):
You should pay attention to the naming.
Branch - "[Ticket]-[Description]" (e.g. EPMLSTRCMW-220-launch-pipeline-docker-compose)
Commit - "[Type]: [Subject in imperative tone]" (e.g. fix: update regex for commit message)

Also, you can read about git commit messages here:  docs/git-commit-message-style-guide.md and https://chris.beams.io/posts/git-commit/
Also, you can install the hook for checking to commit message hooks/install.sh

And last: you should squash commits before the push.

PR #154, opened 2021-04-28
EPMLSTRCMW-214

OlegKoshelev:
Project versioning has been implemented.

- "version" column has been added to project table in DB
- "version" field has been added to Project and LocalProject classes
- Previous tests have been fixed 
- New tests have been added for ProjectRepository.updateProjectVersion  and ProjectService.updateProjectVersion
- changelog-0.1.xml has been updated

### DIFF
--- a/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
+++ b/datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
@@ -65,6 +65,15 @@
         </rollback>
     </changeSet>
 
+    <changeSet id="add_column_for_project_version" author="oleg_koshelev">
+        <addColumn tableName="project">
+            <column name="version" type="varchar(255)" defaultValue="v0.0.1"/>
+        </addColumn>
+        <rollback>
+            <dropColumn tableName="project" columnName="version"/>
+        </rollback>
+    </changeSet>
+
     <changeSet id="create_run_table" author="manuel_vigueras">
         <createTable tableName="run">
             <column name="run_id" type="${uuid.type}"/>
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
@@ -15,6 +15,7 @@ final case class Project(
   name: String,
   active: Boolean,
   repositoryId: RepositoryId,
+  version: PipelineVersion,
   visibility: Visibility = Private
 )
 object Project {
@@ -26,7 +27,8 @@ final case class LocalProject(
   ownerId: UserId,
   name: String,
   active: Boolean,
-  visibility: Visibility = Private
+  visibility: Visibility = Private,
+  version: PipelineVersion
 ) {
   def toProject(repositoryId: RepositoryId): Project =
     Project(
@@ -35,7 +37,8 @@ final case class LocalProject(
       name = name,
       active = active,
       repositoryId = repositoryId,
-      visibility = visibility
+      visibility = visibility,
+      version = version
     )
 }
 
@@ -91,6 +94,7 @@ object GitLabVersion {
 final case class PipelineVersion(major: VersionValue, minor: VersionValue, revision: VersionValue)
     extends Ordered[PipelineVersion] {
   import VersionValue._
+  import PipelineVersion._
   def name: String = s"v$major.$minor.$revision"
 
   private val ordering: Ordering[PipelineVersion] = Ordering.by(v => (v.major, v.minor, v.revision))
@@ -108,6 +112,8 @@ final case class PipelineVersion(major: VersionValue, minor: VersionValue, revis
     this.copy(revision = increment(this.revision))
 
   override def toString: String = this.name
+
+  def value: String = this.name
 }
 
 object PipelineVersion {
--- a/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
+++ b/repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
@@ -20,16 +20,18 @@ object TestProjectUtils {
     name: String = s"project-$randomUuidStr",
     repository: RepositoryId = getDummyRepositoryId,
     active: Boolean = true,
+    version: PipelineVersion = PipelineVersion("v0.0.1"),
     visibility: Visibility = Private
-  ): Project = Project(projectId, ownerId, name, active, repository, visibility)
+  ): Project = Project(projectId, ownerId, name, active, repository, version, visibility)
   def getDummyLocalProject(
     projectId: ProjectId = getDummyProjectId,
     ownerId: UserId = TestUserUtils.getDummyUserId,
     name: String = s"project-$randomUuidStr",
     active: Boolean = true,
+    version: PipelineVersion = PipelineVersion("v0.0.1"),
     visibility: Visibility = Private
   ): LocalProject =
-    LocalProject(projectId, ownerId, name, active, visibility)
+    LocalProject(projectId, ownerId, name, active, visibility, version)
   def getDummyCommit(id: String = randomUuidStr): Commit = Commit(id)
   def getDummyPipeLineVersion(
     v1: Int = 1 + randomInt(),
--- a/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
@@ -50,15 +50,15 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     responseBody: UpdateFiledResponse
   )(
     implicit ec: ExecutionContext
-  ): AsyncResult[UpdateFiledResponse] =
+  ): AsyncResult[(UpdateFiledResponse, PipelineVersion)] =
     createTag(repositoryId, version).map {
-      case Right(_)        => Right(responseBody)
+      case Right(_)        => Right(responseBody, version)
       case Left(exception) => Left(exception)
     }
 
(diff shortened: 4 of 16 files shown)


datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -65,6 +65,15 @@
         </rollback>
     </changeSet>
 
+    <changeSet id="add_column_for_project_version" author="oleg_koshelev">
```
myazinn, line 68: Please put these changes at the bottom. In this case it doesn't make much of a difference, but in general it will affect migration order on a new database with unpredictable consequences.

datasource/src/main/resources/liquibase/changelog/changelog-0.1.xml
```diff
@@ -65,6 +65,15 @@
...
+        <addColumn tableName="project">
+            <column name="version" type="varchar(255)" defaultValue="v0.0.1"/>
```
myazinn, line 70: Looks like `version` is a required field. In this case, you should add `not null` constraint.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -26,7 +27,8 @@ final case class LocalProject(
   ownerId: UserId,
   name: String,
   active: Boolean,
-  visibility: Visibility = Private
+  visibility: Visibility = Private,
+  version: PipelineVersion
```
myazinn, line 31: Always keep parameters with default values at the end for a more convenient use. I.e. place `version` before `visibility`

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -91,6 +94,7 @@ object GitLabVersion {
 final case class PipelineVersion(major: VersionValue, minor: VersionValue, revision: VersionValue)
     extends Ordered[PipelineVersion] {
   import VersionValue._
+  import PipelineVersion._
```
myazinn, line 97: Why do we need this import?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/Project.scala
```diff
@@ -108,6 +112,8 @@ final case class PipelineVersion(major: VersionValue, minor: VersionValue, revis
     this.copy(revision = increment(this.revision))
 
   override def toString: String = this.name
+
+  def value: String = this.name
```
myazinn, line 116: Why do we need this method?

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -20,16 +20,18 @@ object TestProjectUtils {
     name: String = s"project-$randomUuidStr",
     repository: RepositoryId = getDummyRepositoryId,
     active: Boolean = true,
+    version: PipelineVersion = PipelineVersion("v0.0.1"),
```
myazinn, line 23: Use `getDummyPipeLineVersion` instead

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/utils/TestProjectUtils.scala
```diff
@@ -20,16 +20,18 @@ object TestProjectUtils {
...
     visibility: Visibility = Private
-  ): Project = Project(projectId, ownerId, name, active, repository, visibility)
+  ): Project = Project(projectId, ownerId, name, active, repository, version, visibility)
   def getDummyLocalProject(
     projectId: ProjectId = getDummyProjectId,
     ownerId: UserId = TestUserUtils.getDummyUserId,
     name: String = s"project-$randomUuidStr",
     active: Boolean = true,
+    version: PipelineVersion = PipelineVersion("v0.0.1"),
```
myazinn, line 31: Use `getDummyPipeLineVersion` instead

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -50,15 +50,15 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
...
     createTag(repositoryId, version).map {
-      case Right(_)        => Right(responseBody)
+      case Right(_)        => Right(responseBody, version)
       case Left(exception) => Left(exception)
     }
 
   override def updateFile(project: Project, projectFile: ProjectFile, userVersion: Option[PipelineVersion])(
     implicit ec: ExecutionContext
-  ): AsyncResult[UpdateFiledResponse] = {
+  ): AsyncResult[(UpdateFiledResponse, PipelineVersion)] = {
```
myazinn, line 61: It is ok to sometimes use tuples internally. But you should avoid using tuples in public methods. Let's create a new case class that will model your response. It will resolve two problems
1) Tuples are hard two read. What is `PipelineVersion` in this case? Is it new version? Is it previous version? Is it current version?
2) Tuples are not scaling well. If we add a new field to a returned value, most likely we will have to go through a lot of usages of such method and modify them, because return type would be different.
Once you add this class, don't forget to change the return type of `handleCreateTag` as well.

services/src/main/scala/cromwell/pipeline/service/GitLabProjectVersioning.scala
```diff
@@ -190,6 +190,10 @@ class GitLabProjectVersioning(httpClient: HttpClient, config: GitLabConfig)
     implicit ec: ExecutionContext
   ): AsyncResult[List[GitLabVersion]] = ???
 
+  override def getDefaultProjectVersion()(
+    implicit ec: ExecutionContext
+  ): PipelineVersion = PipelineVersion(config.defaultFileVersion)
```
myazinn, line 195: Re-use this method in other places in `GitLabProjectVersioning`

services/src/main/scala/cromwell/pipeline/service/ProjectFileService.scala
```diff
@@ -28,7 +28,16 @@ class ProjectFileService(
   ): Future[Either[VersioningException, UpdateFiledResponse]] =
     projectService
       .getUserProjectById(projectId, userId)
-      .flatMap(project => projectVersioning.updateFile(project, projectFile, version))
+      .flatMap(
```
myazinn, line 31:
```suggestion
      .flatMap {
```
Is more convenient to use curly braces in case of multiline statements. Besides, you already use curly braces on line 33, so it would be more consistent.

services/src/main/scala/cromwell/pipeline/service/ProjectFileService.scala
```diff
@@ -28,7 +28,16 @@ class ProjectFileService(
...
+        project =>
+          projectVersioning.updateFile(project, projectFile, version).map {
+            case Right((response, version)) => {
```
myazinn, line 34:
```suggestion
            case Right((response, version)) =>
```
redundant braces

services/src/main/scala/cromwell/pipeline/service/ProjectFileService.scala
```diff
@@ -28,7 +28,16 @@ class ProjectFileService(
...
+              projectService.updateProjectVersion(version, projectId, userId)
+              Right(response)
```
myazinn, line 36: `projectService.updateProjectVersion` returns `Future`. Which means it's possible that this future will fail, but the client will never know about it and will receive successful response. Here's what you should do:
1) Write a test that reproduces this case
2) Make sure it fails
3) Fix `ProjectFileService`
4) Make sure test works

services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
```diff
@@ -48,6 +47,22 @@ class ProjectServiceTest extends AsyncWordSpec with Matchers with MockitoSugar w
       }
     }
 
+    "updateProjectVersion" should {
+      "return correct version of project" taggedAs Service in {
+        val projectId = ProjectId("projectId")
+        val userId = UserId.random
+        val version = PipelineVersion("v1.0.0")
+
+        when(projectRepository.getProjectById(projectId)).thenReturn(Future(Some(dummyProject.copy(version = version))))
```
myazinn, line 56:
```suggestion
        when(projectRepository.getProjectById(projectId)).thenReturn(Future.successful(Some(dummyProject.copy(version = version))))
```

myazinn: Use `Future.successful` for fast operations that don't require asynchronous execution
https://stackoverflow.com/questions/21466652/what-is-the-difference-between-future-successfulnone-and-futurenone

services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
```diff
@@ -48,6 +47,22 @@ class ProjectServiceTest extends AsyncWordSpec with Matchers with MockitoSugar w
...
+
+        projectService.updateProjectVersion(version, projectId, userId)
+        projectRepository.getProjectById(projectId).map {
+          case Some(project) => project.version shouldBe version
+          case None          => fail()
```
myazinn, line 61: This test tests nothing.
`projectRepository` is a mock that you set to return specific value. And here you are directly calling this mock to check that mock returns exactly what you set it to return. No `ProjectService` testing here.
If you modify `ProjectService.updateProjectVersion` to fail on any request, then this test will pass regardless.

myazinn (changes requested):
Content of this PR does not correspond to a linked ticket.
Title of the linked ticket is `There is no versioning for the ProjectFileConfiguration` and description says that we should add it. I don't see how changes in this PR resolve it. Maybe you wanted to link another ticket?
Anyway, I left some comments on your changes.

PR #160, opened 2021-05-12
EPMLSTRCMW-200 fix: Stop application in case of initialization error

gominasian:
Application stops when initialization error is met
cromwell.pipeline excluded from coverage

### DIFF
--- a/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
+++ b/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
@@ -11,25 +11,14 @@ import cromwell.pipeline.utils.configs.ConfigJsonOps
 import org.slf4j.LoggerFactory
 
 import scala.concurrent.ExecutionContextExecutor
+import scala.util.{ Failure, Success, Try }
 
 object CromwellPipelineApp extends App {
 
+  val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
+
   implicit val system: ActorSystem = ActorSystem("cromwell-pipeline")
   implicit val executionContext: ExecutionContextExecutor = system.dispatcher
-
-  val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
-  val components = new ApplicationComponents()
-
-  import components.applicationConfig
-  import components.applicationConfig.webServiceConfig
-  import components.controllerModule._
-  import components.datastorageModule._
-
-  pipelineDatabaseEngine.updateSchema()
-
-  type SecuredRoute = AccessTokenContent => Route
-  def routeCombiner(routes: SecuredRoute*): SecuredRoute = token => concat(routes.map(_(token)): _*)
-
   implicit val rejectionHandler: RejectionHandler = RejectionHandler
     .newBuilder()
     .handle {
@@ -39,17 +28,38 @@ object CromwellPipelineApp extends App {
     .result()
     .withFallback(RejectionHandler.default)
 
-  val route = authController.route ~ securityDirective.authenticated {
-    routeCombiner(
-      userController.route,
-      projectController.route,
-      projectFileController.route,
-      runController.route,
-      configurationController.route
-    )
-  }
+  type SecuredRoute = AccessTokenContent => Route
 
-  log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
-  log.info(ConfigJsonOps.configToJsonString(applicationConfig))
-  Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
+  def routeCombiner(routes: SecuredRoute*): SecuredRoute = token => concat(routes.map(_(token)): _*)
+
+  Try {
+    val components = new ApplicationComponents()
+
+    import components.applicationConfig
+    import components.applicationConfig.webServiceConfig
+    import components.controllerModule._
+    import components.datastorageModule._
+
+    pipelineDatabaseEngine.updateSchema()
+
+    val route = authController.route ~ securityDirective.authenticated {
+      routeCombiner(
+        userController.route,
+        projectController.route,
+        projectFileController.route,
+        runController.route,
+        configurationController.route
+      )
+    }
+
+    log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
+    log.info(ConfigJsonOps.configToJsonString(applicationConfig))
+    Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
+  } match {
+    case Success(_) =>
+      log.info(s"Application is running!")
+    case Failure(e) =>
+      system.terminate()
+      log.error(s"Application cannot be started:", e)
+  }
 }


OlegKoshelev:
Add a description to your task, please

DruzhininPavel (approved):
No questions.

build.sbt
```diff
@@ -28,7 +28,8 @@ lazy val commonSettings = Seq(
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-h", "target/test-reports"),
   coverageEnabled in Test := true,
   coverageMinimum := 60,
-  coverageFailOnMinimum := true
+  coverageFailOnMinimum := true,
+  coverageExcludedPackages := "CromwellPipelineApp\\..*"
```
Nikolai-Isaev: Maybe we should put more entities here :)

Nikolai-Isaev (approved):

build.sbt
myazinn: Definitely. We should exclude from coverage everything in `portal` module.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -37,16 +38,26 @@ object CromwellPipelineApp extends App {
...
-    routeCombiner(
-      userController.route,
-      projectController.route,
-      projectFileController.route,
-      runController.route,
-      configurationController.route
-    )
-  }
+  try {
+    val route = authController.route ~ securityDirective.authenticated {
```
myazinn: By this moment application could have faced tons of other errors. Maybe we couldn't create some service, or maybe database migration failed.
Basically every step that comes after ActorSystem creation must be treated as unsafe step, i.e. included in try-catch statement.

myazinn (changes requested):

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
myazinn: @gominasian probably you didn't notice it

build.sbt
```diff
@@ -28,7 +28,8 @@ lazy val commonSettings = Seq(
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-h", "target/test-reports"),
   coverageEnabled in Test := true,
   coverageMinimum := 60,
-  coverageFailOnMinimum := true
+  coverageFailOnMinimum := true,
+  coverageExcludedPackages := "<empty>;cromwell\\.pipeline.*"
```
myazinn: Something is wrong. I don't see a lot of module here
https://coveralls.io/builds/39858444
Compare this to another report
https://coveralls.io/builds/39810273

myazinn (changes requested):

DruzhininPavel (approved):
Ok. Now it is work.

iliaonishchenko:
Is changes request fixed?
Can I review it? MR is in draft status.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,58 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
+
   implicit val system: ActorSystem = ActorSystem("cromwell-pipeline")
   implicit val executionContext: ExecutionContextExecutor = system.dispatcher
 
-  val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
   val components = new ApplicationComponents()
 
   import components.applicationConfig.webServiceConfig
   import components.controllerModule._
   import components.datastorageModule._
```
iliaonishchenko: Why not to move these lines inside `Try`?

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,58 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
-
-  val route = authController.route ~ securityDirective.authenticated {
-    routeCombiner(
-      userController.route,
-      projectController.route,
-      projectFileController.route,
-      runController.route,
-      configurationController.route
-    )
+  val isSuccessful = Try {
```
iliaonishchenko: Should we have such value at all?
We can do matching at `Try` statement.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,58 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
   }
 
-  log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
-  Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
+  isSuccessful match {
+    case Success(_) =>
+      log.info(s"Application is running")
+    case Failure(e) =>
+      system.terminate()
+      log.info(s"Application cannot be started.\n${e.fillInStackTrace()}")
```
iliaonishchenko: When you log error, it's probably better to use `error` level.

iliaonishchenko (changes requested):
Nice job, there're some comments:

OlegKoshelev (approved):

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
gominasian: Thank you, will fix it

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
gominasian: Fixed it

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
gominasian: Done

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
gominasian: Yep, fixed

GrigoriiBerezin (approved):
Nice

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,56 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
+
+    val route = Route.seal(authController.route ~ securityDirective.authenticated {
+      routeCombiner(
+        userController.route,
+        projectController.route,
+        projectFileController.route,
+        runController.route,
+        configurationController.route
+      )
+    })(rejectionHandler = rejectionHandler)
```
Dmitrii-Shevchenko: Why do you signify rejectionHandler explicitly?

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,56 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
+    implicit val rejectionHandler: RejectionHandler = RejectionHandler
+      .newBuilder()
+      .handle {
+        case ValidationRejection(msg, _)      => complete(StatusCodes.BadRequest, msg)
+        case MissingAccessTokenRejection(msg) => complete(StatusCodes.Unauthorized, msg)
+      }
+      .result()
+      .withFallback(RejectionHandler.default)
+
+    val route = Route.seal(authController.route ~ securityDirective.authenticated {
```
Dmitrii-Shevchenko: What is the reason that you add routes via seal?

Dmitrii-Shevchenko:
Squash commits please

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,43 +10,56 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
+
+    log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
+    Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
+
+  } match {
+    case Success(_) =>
+      log.info(s"Application is running")
+    case Failure(e) =>
+      system.terminate()
+      log.error(s"Application cannot be started.\n${e.fillInStackTrace()}")
```
iliaonishchenko: When you log at `error` level, you provide text description and an error as a second argument. No need to interpolate Exception to string.

build.sbt
```diff
@@ -29,7 +29,7 @@ lazy val commonSettings = Seq(
   coverageEnabled in Test := true,
   coverageMinimum := 60,
   coverageFailOnMinimum := true,
-  coverageExcludedPackages := "cromwell.pipeline.database.*"
+  coverageExcludedPackages := "<empty>; cromwell\\.pipeline; cromwell\\.pipeline\\.database.*; "
```
iliaonishchenko: Could you please describe what it does?

iliaonishchenko: Why there is `empty` in front of the string?

gominasian: This are regex to exclude some files from coverage:
I used escape slashes (\\) to make dots match literally.

- empty tag uses for pick up first exclusion. To be honest, it works perfect without it too, but I think there are cases when you really need this tag. And it's in examples in scoverage docs but without any description.

- cromwell\\.pipeline excludes files straight from cromwell.pipeline, but not from cromwell.pipeline.database, for example

- cromwell\\.pipeline\\.database.* excludes everything from cromwell.pipeline.database and inside packages (but it is from some other commit, I just put escape slashes there)

iliaonishchenko:
Build fails.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -10,24 +10,14 @@ import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
...
 
 import scala.concurrent.ExecutionContextExecutor
+import scala.util.{ Failure, Success, Try }
 
 object CromwellPipelineApp extends App {
 
-  implicit val system: ActorSystem = ActorSystem("cromwell-pipeline")
-  implicit val executionContext: ExecutionContextExecutor = system.dispatcher
-
   val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
```
Dmitrii-Shevchenko, line 18: It would be better if you keep the previous order

gominasian: Which one?

Dmitrii-Shevchenko: For example variable `log` and implicites.
You replaced the lines in order, but that doesn't make any sense.

gominasian: Is it important to keep previous order? I've put log before implicits, bc I want them to stay together with rejectionHandler, but it's a big block of code and logger could be inconspicuous this way

Dmitrii-Shevchenko (approved):
Good work!

iliaonishchenko (changes requested):
Nice work! I'm ok with all changes, the only thing you need to do is to change commit message according to the project's guidelines.

iliaonishchenko (approved):
Looks good to me!

project/Dependencies.scala
```diff
@@ -58,6 +59,7 @@ object Dependencies {
   val logback = "ch.qos.logback" % "logback-classic" % Version.logback
   val wireMock = "com.github.tomakehurst" % "wiremock" % Version.wireMock % Test
   val sl4j = "org.slf4j" % "slf4j-api" % Version.sl4j
+  val slf4j = "org.slf4j" % "slf4j-api" % Version.slf4j
```
myazinn: Why do we need the same dependency twice?

myazinn (commented):
Main changes look good to me. Although I have a question about new dependency.

project/Dependencies.scala
gominasian: Fix it

myazinn (approved):

PR #166, opened 2021-05-18
EPMLSTRCMW-185 Feat Print all configuration parameters during the application initialization 

Nikolai-Isaev:
feat: LogInfoUtilsPlayJson created - print config message to console using Play Json lib

### DIFF
--- a/utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala
+++ b/utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala
@@ -31,6 +31,14 @@ final case class MongoConfig(
   collection: String
 ) extends ConfigComponent
 
+final case class PostgreConfig(
+  serverName: String,
+  portNumber: Int,
+  databaseName: String,
+  user: String,
+  password: Array[Char]
+) extends ConfigComponent
+
 final case class ExpirationTimeInSeconds(accessToken: Long, refreshToken: Long, userSession: Long)
 
 class ApplicationConfig(val config: Config) {
@@ -88,9 +96,25 @@ class ApplicationConfig(val config: Config) {
     )
   }
 
+  lazy val postgreConfig: PostgreConfig = {
+    val _config = config.getConfig("database.postgres_dc.db.properties")
+    PostgreConfig(
+      serverName = _config.getString("serverName"),
+      portNumber = _config.getInt("portNumber"),
+      databaseName = _config.getString("databaseName"),
+      user = _config.getString("user"),
+      password = _config.getString("password").toCharArray
+    )
+  }
+
 }
 
 object ApplicationConfig {
   def load(config: Config = ConfigFactory.load()): ApplicationConfig =
     new ApplicationConfig(config)
+
+  def unapply(
+    config: ApplicationConfig
+  ): Option[(WebServiceConfig, AuthConfig, GitLabConfig, MongoConfig, PostgreConfig)] =
+    Option((config.webServiceConfig, config.authConfig, config.gitLabConfig, config.mongoConfig, config.postgreConfig))
 }
--- a/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
+++ b/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
@@ -7,6 +7,7 @@ import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.{ RejectionHandler, Route, ValidationRejection }
 import cromwell.pipeline.auth.token.MissingAccessTokenRejection
 import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.utils.configs.ConfigJsonOps
 import org.slf4j.LoggerFactory
 
 import scala.concurrent.ExecutionContextExecutor
@@ -19,6 +20,7 @@ object CromwellPipelineApp extends App {
   val log = LoggerFactory.getLogger(CromwellPipelineApp.getClass)
   val components = new ApplicationComponents()
 
+  import components.applicationConfig
   import components.applicationConfig.webServiceConfig
   import components.controllerModule._
   import components.datastorageModule._
@@ -48,5 +50,6 @@ object CromwellPipelineApp extends App {
   }
 
   log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
+  log.info(ConfigJsonOps.configToJsonString(applicationConfig))
   Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
 }
--- a/utils/src/main/scala/cromwell/pipeline/utils/configs/ConfigJsonOps.scala
+++ b/utils/src/main/scala/cromwell/pipeline/utils/configs/ConfigJsonOps.scala
@@ -1,0 +1,84 @@
+package cromwell.pipeline.utils.configs
+
+import cromwell.pipeline.utils._
+import cromwell.pipeline.utils.configs.SecretData._
+import play.api.libs.functional.syntax.{ toFunctionalBuilderOps, unlift }
+import play.api.libs.json.Format.GenericFormat
+import play.api.libs.json.{ Json, Writes, __ }
+
+trait ConfigJsonOps {
+  def configToJsonString(applicationConfig: ApplicationConfig): String
+}
+
+object ConfigJsonOps extends ConfigJsonOps {
+
+  def configToJsonString(applicationConfig: ApplicationConfig): String = {
+
+    val secretCAWrites: Writes[SecretData[Array[Char]]] = secretDataWrites(Writes.StringWrites.contramap(new String(_)))
+
+    implicit val wsWrites: Writes[WebServiceConfig] =
+      ((__ \ "interface").write[String] ~
+        (__ \ "port").write[Int])(unlift(WebServiceConfig.unapply))
+
+    implicit val expTimeWrites: Writes[ExpirationTimeInSeconds] =
+      ((__ \ "accessToken").write[Long] ~
+        (__ \ "refreshToken").write[Long] ~
+        (__ \ "userSession").write[Long])(unlift(ExpirationTimeInSeconds.unapply))
+
+    implicit val authWrites: Writes[AuthConfig] =
+      ((__ \ "secretKey").write[SecretData[String]] ~
+        (__ \ "hmacAlgorithm").write[String] ~
+        (__ \ "expirationTimeInSeconds").write[ExpirationTimeInSeconds])(
+        authConfig =>
+          unlift(AuthConfig.unapply)(authConfig).copy(_1 = authConfig.secretKey, _2 = authConfig.hmacAlgorithm.toString)
+      )
+
+    implicit val gitLabWrites: Writes[GitLabConfig] =
+      ((__ \ "url").write[String] ~
+        (__ \ "token").write[SecretData[Map[String, String]]] ~
+        (__ \ "defaultFileVersion").write[String] ~
+        (__ \ "defaultBranch").write[String])(
+        gitLabConfig => unlift(GitLabConfig.unapply)(gitLabConfig).copy(_2 = gitLabConfig.token)
+      )
+
+    implicit val mongoWrites: Writes[MongoConfig] =
+      ((__ \ "user").write[String] ~
+        (__ \ "password").write[SecretData[Array[Char]]](secretCAWrites) ~
+        (__ \ "host").write[String] ~
+        (__ \ "port").write[Int] ~
+        (__ \ "authenticationDatabase").write[String] ~
+        (__ \ "database").write[String] ~
+        (__ \ "collection").write[String])(
+        mongoConfig => unlift(MongoConfig.unapply)(mongoConfig).copy(_2 = mongoConfig.password)
+      )
+
+    implicit val postgreWrites: Writes[PostgreConfig] =
+      ((__ \ "serverName").write[String] ~
+        (__ \ "portNumber").write[Int] ~
+        (__ \ "databaseName").write[String] ~
+        (__ \ "user").write[String] ~
+        (__ \ "password").write[SecretData[Array[Char]]](secretCAWrites))(
+        postgreConfig => unlift(PostgreConfig.unapply)(postgreConfig).copy(_5 = postgreConfig.password)
+      )
+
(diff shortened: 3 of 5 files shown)


utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala
```diff
@@ -88,9 +96,25 @@ class ApplicationConfig(val config: Config) {
...
+
 }
 
 object ApplicationConfig {
   def load(config: Config = ConfigFactory.load()): ApplicationConfig =
     new ApplicationConfig(config)
+
+  def unapply(
+    config: ApplicationConfig
+  ): Option[(WebServiceConfig, AuthConfig, GitLabConfig, MongoConfig, PostgreConfig)] =
```
DruzhininPavel, line 118: I think, you can to use some trait like SomeConfig or different name for combine all configs under one type. It shod help if you want to add a new config.

DruzhininPavel (approved):
Have no questions

utils/src/main/scala/cromwell/pipeline/utils/ApplicationConfig.scala
Nikolai-Isaev: Great idea. Can you be more specific? Can we discuss it in teams?

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+        (__ \ "AuthConfig").write[AuthConfig] ~
+        (__ \ "GitLabConfig").write[GitLabConfig] ~
+        (__ \ "MongoConfig").write[MongoConfig] ~
+        (__ \ "PostgreConfig").write[PostgreConfig])(unlift(ApplicationConfig.unapply))
+
+    val messageHeader =
+      "\n---------------------------------------------------\n" +
+        "Application starts with the following configuration\n" +
+        "---------------------------------------------------\n"
+
```
gominasian: May be you should change it for smth like that:
  val message = "Application starts with the following configuration"
  val delimiterLine = "-" * messageHeader.length
  val messageHeader = s"$delimiterLine\n$message\n$delimiterLine"
It would keep you from counting letters if you would change message.
but it's just for my taste

utils/src/test/scala/cromwell/pipeline/utils/configs/LogInfoUtilsTest.scala
```diff
@@ -0,0 +1,114 @@
...
+       |    user = "testMongoUser"
+       |    password = "testMongoPassword"
+       |    host = "testMongoHost"
+       |    port = 1
+       |    authenticationDatabase = "testAuthDatabase"
+       |    database = "testMongoDatabase"
+       |    collection = "testMongoCollection"
+       |  }
+       |}""".stripMargin
+
```
gominasian: Is it necessary to check the whole thing? Or you can test just some lines?

gominasian (approved):
I have some minor questions about this PR, but everything else is ok for me

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Great idea, actually. The less literals are - the better. It wasn't my motto for this task, unfortunately :)

utils/src/test/scala/cromwell/pipeline/utils/configs/LogInfoUtilsTest.scala
Nikolai-Isaev: Thats a good question. The function tested is strictly tied with our current configuration structure, and I couldn't find a way to avoid creating full-size dummies.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Fixed with adding the createHeader method

utils/src/test/scala/cromwell/pipeline/utils/configs/LogInfoUtilsTest.scala
myazinn: Honestly, it's an implementation flaw :) We should be able to pretty print every single configuration class separately. For now I think it's ok but keep in mind that on the real project most likely you would have to make testable every config separately
Of course it makes sense to test them as a whole too.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+trait LogInfoUtils {
+  def getConfigMessage: String
+}
+
+object LogInfoUtils {
+
+  def getConfigMessage(applicationConfig: ApplicationConfig): String = {
+
+    def createMessageHeader: String = {
```
Dmitrii-Shevchenko: Do you need to calculate this every time? I think val is better here

Nikolai-Isaev: Great point, ill fix it, thanks

utils/src/test/scala/cromwell/pipeline/utils/configs/LogInfoUtilsTest.scala
Nikolai-Isaev: Maybe I should return to that issue in a month then and see if I can do it better with a new experience :)

Dmitrii-Shevchenko (approved):

OlegKoshelev (approved):

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import cromwell.pipeline.utils._
+import cromwell.pipeline.utils.configs.SecretData._
+import play.api.libs.functional.syntax.{ toFunctionalBuilderOps, unlift }
+import play.api.libs.json.Format.GenericFormat
+import play.api.libs.json.{ Json, Writes, __ }
+
+trait LogInfoUtils {
+  def getConfigMessage: String
+}
```
iliaonishchenko: Why do you need this trait?

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+
+object LogInfoUtils {
```
iliaonishchenko: How this object is related to the Logging?

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  def getConfigMessage(applicationConfig: ApplicationConfig): String = {
```
iliaonishchenko: From my side, method's name is confusing. We can come up with the name corresponding to the action performed by the method.
How about `configToJson` or `configToJsonString`?

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+      ((__ \ "WebServiceConfig").write[WebServiceConfig] ~
+        (__ \ "AuthConfig").write[AuthConfig] ~
+        (__ \ "GitLabConfig").write[GitLabConfig] ~
+        (__ \ "MongoConfig").write[MongoConfig] ~
+        (__ \ "PostgreConfig").write[PostgreConfig])(unlift(ApplicationConfig.unapply))
+
+    val messageHeader: String = {
+      val headerDelimiter = "-"
+      val headerMessage = "Application starts with the following configuration"
+      val headerDelimeterLine = s"${headerDelimiter * headerMessage.length}"
```
iliaonishchenko: Typo.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
```diff
@@ -0,0 +1,84 @@
...
+    implicit val appConfigWrites: Writes[ApplicationConfig] =
+      ((__ \ "WebServiceConfig").write[WebServiceConfig] ~
+        (__ \ "AuthConfig").write[AuthConfig] ~
+        (__ \ "GitLabConfig").write[GitLabConfig] ~
+        (__ \ "MongoConfig").write[MongoConfig] ~
+        (__ \ "PostgreConfig").write[PostgreConfig])(unlift(ApplicationConfig.unapply))
+
+    val messageHeader: String = {
+      val headerDelimiter = "-"
+      val headerMessage = "Application starts with the following configuration"
```
iliaonishchenko: Could you please add a colon at the end of the string? :)

iliaonishchenko (changes requested):
Good work, well done!
I have a few comments:
- General:
  - Please double check how to write commit messages.
  - Looks like you've been committing on behalf of two git users :) Could you please fix it?

- Code-related:

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Something like "ConfigsToStringUtils" would definitely be a better option, thanks!

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: As far as I understand from our mentors explanation this may be useful for future functionality expansion and may be a good practice in general. Quite possible I got it wrong though.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Fixed

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Fixed

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Done!

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
iliaonishchenko: Having `Utils` at the end still looks very much as Java naming.. :(
Please reconsider it.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
iliaonishchenko: Well, it might be, but you need to use it. 
Right now if you remove this trait it will change nothing. It won't even request any code changes.
We should not have anything we're not using in our codebase.

iliaonishchenko:
@Nikolai-Isaev, for the future notice:
The person who started conversation should mark it as resolved, not you.
Other way it might be more complicated to understand change request status.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: So that object is designed to collect configurations and make a nice readable string out of it. The name should encompass those properties, right?
How about: ConfigsPrinter, ConfigsToStringTransformer, ConfigsToString, ConfigsToJsonString, ConfigsMessenger, ConfigsMessageGenerator, ConfigsPrettyPrinter 
Any advice on proper naming would be really appreciated.

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
myazinn: @ilyaonishenko @Nikolai-Isaev 
Let me explain why asked to add this trait.
It's quite hard to test classes that use classic `Utils` objects. I.e.
```
object Utils {
  def foo(n: Int): Unit = println(n)
}
class A {
  def hardToTest(str: String): Int = {
    val i = str.toInt
    Utils.foo(i)
    i
  }
}
```
It's tricky to verify that `hardToTest` called `foo` method if we want to write unit test.
When we use utils objetcs, we bypass DI and rely on some global state. It is no better than
```
class Utils {
  def foo(n: Int): Unit = println(n)
}
class A {
  private val utils = new Utils
  def hardToTest(str: String): Int = {
    val i = str.toInt
    utils.foo(i)
    i
  }
}
```
That's why I asked to add a separate trait for this `utils` class. If we use trait, then it will be easily testable and reusable
```
trait Utils {
  def foo(n: Int): Unit
}
object Utils extends Utils {
  def foo(n: Int): Unit = println(n)
}
class A(utils: Utils) {
  def easyToTest(str: String): Int = {
    val i = str.toInt
    utils.foo(i)
    i
  }
}
```
And we still can use it as classic `Utils` class if required.
In this particular case it won't make much of a difference, but it might be useful in the future and is a good starting point for getting rid of classic `utils`. You are free to drop it if it casuses too many problems.

iliaonishchenko: @myazinn, that's could be a case, if `LogInfoUtils` extends this trait. :)
But right now this trait just exists nearby. 
@Nikolai-Isaev, you need to make LogInfoUtils to use this trait.

Nikolai-Isaev: Done

utils/src/main/scala/cromwell/pipeline/utils/configs/LogInfoUtils.scala
Nikolai-Isaev: Renamed to ConfigJsonOps

iliaonishchenko (approved):
Looks good to me :)
Nice work.

PR #170, opened 2021-05-27
EPMLSTRCMW-222

OlegKoshelev:
**Fix migrate to codecov.**

- cmwl_pipeline_ci.yml,  build.sbt and project/plugins.sbt  have been updated.
-  codecov.yml has been added.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -27,9 +27,10 @@ lazy val commonSettings = Seq(
   test in Test := (test in Test).dependsOn(checkFormat).value,
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-h", "target/test-reports"),
   coverageEnabled in Test := true,
-  coverageMinimum := 60,
+  coverageMinimum := 50,
   coverageFailOnMinimum := true,
-  coverageExcludedPackages := "cromwell.pipeline.database.*"
+  coverageExcludedPackages :=
+    "cromwell\\.pipeline; cromwell\\.pipeline\\.database.*"
 )
 
 addCommandAlias("fmt", "all scalafmtSbt scalafmt test:scalafmt")
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0%
+        informational: true
--- a/.github/workflows/cmwl_pipeline_ci.yml
+++ b/.github/workflows/cmwl_pipeline_ci.yml
@@ -29,7 +29,7 @@ jobs:
         run: sbt clean compile undeclaredCompileDependenciesTest
 
       - name: Run tests
-        run: sbt coverage test coverage it:test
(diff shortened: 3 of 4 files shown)


build.sbt
```diff
@@ -28,7 +28,7 @@ lazy val commonSettings = Seq(
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-h", "target/test-reports"),
   coverageEnabled in Test := true,
   coverageMinimum := 60,
-  coverageFailOnMinimum := true,
+  coverageFailOnMinimum := false,
```
OlegKoshelev: @ilyaonishenko Now seems it works but I'm not sure about this line. If use "true"  the tests won't pass due to coverage  <  60% ( you can see it during run test job). Maybe you can suggest a better solution?

Nikolai-Isaev (approved):
It works, nice job! Will this bot always comment on coverage here btw?

DruzhininPavel (approved):
Yep. it is normal!

OlegKoshelev:
> It works, nice job! Will this bot always comment on coverage here btw?
i think so yes

gominasian (approved):

build.sbt
```diff
@@ -27,9 +27,10 @@ lazy val commonSettings = Seq(
   test in Test := (test in Test).dependsOn(checkFormat).value,
   testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-h", "target/test-reports"),
   coverageEnabled in Test := true,
-  coverageMinimum := 60,
+  coverageMinimum := 50,
```
iliaonishchenko, line 30: Does it work with coverage minimum = 60?

codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        treshold: 0%
```
iliaonishchenko: What threshold stands for?

iliaonishchenko (commented):
Nice work, I have some questions:

P.S. also please fix commit message. 'Migrate to codecov[2]' sounds better then 'Fix migrate to codecov'

build.sbt
OlegKoshelev: Thanks for the clarification! It seems possible. I'll check, but if it doesn't work, I'll leave 50

codecov.yml
OlegKoshelev: As I understood it allows to monitor coverage. If next MR have coverage less it won't pass. For example if threshold = 10% and current coverage 70%, it allows to reduce coverage at next MR to 60%. I think I made a typo, I'll fix it.

build.sbt
OlegKoshelev: @ilyaonishenko Seems it doesn't work with 60% 
(Coverage is below minimum [51.65% < 60.00%]: Stmt:Total
Error:  (womtool / coverageReport) Coverage minimum was not reached)

GrigoriiBerezin (approved):
Nice

PR #171, opened 2021-06-02
EPMLSTRCMW-249

OlegKoshelev:
**DESCRIPTION**

Right now ProjectRepository.getProjectByName returns option of single project
Theoretically, it's possible that several projects may have the same name (or the same name pattern), and the first found project would be authored by another user (not the one who called the method), and as a result user won't get the project (even he is an author of one of such projects).

**CHANGES**
- ProjectEntry.scala, ProjectRepository.scala, ProjectService.scala, ProjectControllerItTest.scala and ProjectServiceTest.scala have been updated.
- Now a query to the database gives out a sequence of projects, it allows to get a project by name with current userId. Therefore we can avoid  getting of project with random (inappropriate) userId at database level.

### DIFF
--- a/portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
+++ b/portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
@@ -48,7 +48,7 @@ class ProjectControllerItTest
 
   "ProjectController" when {
     "getProjectByName" should {
-      "return a project with the same name" in {
+      "return a project with the same name and current userId" in {
         val projectByNameRequest = dummyProject.name
         val accessToken = AccessTokenContent(dummyProject.ownerId)
 
@@ -64,6 +64,16 @@ class ProjectControllerItTest
 
         Get("/projects?name=" + projectByNameRequest) ~> projectController.route(accessToken) ~> check {
           status shouldBe StatusCodes.Forbidden
+        }
+      }
+
+      "return a status code 404 if project doesn't exist" in {
+        val nonExistProject = TestProjectUtils.getDummyProject()
+        val projectByNameRequest = nonExistProject.name
+        val accessToken = AccessTokenContent(nonExistProject.ownerId)
+
+        Get("/projects?name=" + projectByNameRequest) ~> projectController.route(accessToken) ~> check {
+          status shouldBe StatusCodes.NotFound
         }
       }
     }
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
@@ -1,12 +1,11 @@
 package cromwell.pipeline.service
 
-import java.util.UUID
-
 import cromwell.pipeline.datastorage.dao.repository.ProjectRepository
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
 import cromwell.pipeline.service.Exceptions.{ ProjectAccessDeniedException, ProjectNotFoundException }
 
+import java.util.UUID
 import scala.concurrent.{ ExecutionContext, Future }
 
 class ProjectService(projectRepository: ProjectRepository, projectVersioning: ProjectVersioning[VersioningException])(
@@ -14,16 +13,16 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
 ) {
 
   private[service] def getUserProjectById(projectId: ProjectId, userId: UserId): Future[Project] =
-    projectRepository.getProjectById(projectId).flatMap(getForUserOrFail(_, userId))
+    projectRepository.getProjectById(projectId).flatMap(project => getForUserOrFail(project.toSeq, userId))
 
   def getUserProjectByName(namePattern: String, userId: UserId): Future[Project] =
-    projectRepository.getProjectByName(namePattern).flatMap(getForUserOrFail(_, userId))
+    projectRepository.getProjectsByName(namePattern).flatMap(getForUserOrFail(_, userId))
 
-  private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
-    project match {
-      case Some(project) if project.ownerId == userId => Future.successful(project)
-      case Some(_)                                    => Future.failed(new ProjectAccessDeniedException)
-      case None                                       => Future.failed(new ProjectNotFoundException)
+  private def getForUserOrFail(projects: Seq[Project], userId: UserId): Future[Project] =
+    projects.find(_.ownerId == userId) match {
+      case Some(project)             => Future.successful(project)
+      case None if projects.nonEmpty => Future.failed(new ProjectAccessDeniedException)
+      case _                         => Future.failed(new ProjectNotFoundException)
     }
 
   def addProject(request: ProjectAdditionRequest, userId: UserId): Future[Either[VersioningException, Project]] = {
--- a/services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
@@ -4,7 +4,7 @@ import cromwell.pipeline.datastorage.dao.repository.ProjectRepository
 import cromwell.pipeline.datastorage.dao.utils.TestProjectUtils
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
-import cromwell.pipeline.service.Exceptions.ProjectNotFoundException
+import cromwell.pipeline.service.Exceptions.{ ProjectAccessDeniedException, ProjectNotFoundException }
 import org.mockito.Matchers.any
 import org.mockito.Mockito.when
 import org.scalatest.{ AsyncWordSpec, Matchers }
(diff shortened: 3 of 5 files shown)


services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
```diff
@@ -108,5 +108,28 @@ class ProjectServiceTest extends AsyncWordSpec with Matchers with MockitoSugar {
...
+    "getUserProjectByName" should {
+      val project1 = TestProjectUtils.getDummyProject()
+      val project2 = TestProjectUtils.getDummyProject(name = project1.name)
+
+      "return project with corresponding user id" taggedAs Service in {
+        when(projectRepository.getProjectsByName(project2.name)).thenReturn(Future.successful(Seq(project1, project2)))
+        projectService.getUserProjectByName(project2.name, project2.ownerId).map { _ shouldBe project2 }
+      }
+
+      "return project by name" taggedAs Service in {
```
DruzhininPavel: I think at this place we shold write `return project**s** by name`

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -17,7 +17,15 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
...
+      .flatMap(
+        projects =>
+          projects.find(_.ownerId == userId).orElse(projects.headOption) match {
+            case Some(project) => Future.successful(project)
+            case None          => Future.failed(new ProjectNotFoundException)
+          }
+      )
 
   private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
     project match {
```
DruzhininPavel: If we don`t use getForUserOrFail, we should to remowe this method. I think.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
```diff
@@ -58,12 +58,12 @@ class ProjectControllerItTest
         }
       }
 
-      "return status code 403 if user isn`t project owner" in {
+      "return a project with the same name and a random userId" in {
```
DruzhininPavel: What do we recive if project will only for stranger user ?

services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
```diff
@@ -108,5 +108,28 @@ class ProjectServiceTest extends AsyncWordSpec with Matchers with MockitoSugar {
...
       }
     }
+
+    "getUserProjectByName" should {
+      val project1 = TestProjectUtils.getDummyProject()
+      val project2 = TestProjectUtils.getDummyProject(name = project1.name)
+
+      "return project with corresponding user id" taggedAs Service in {
+        when(projectRepository.getProjectsByName(project2.name)).thenReturn(Future.successful(Seq(project1, project2)))
+        projectService.getUserProjectByName(project2.name, project2.ownerId).map { _ shouldBe project2 }
```
DruzhininPavel, line 116: Why do we get only one project, if we recive sequence of two projects ?

services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
OlegKoshelev: Thanks for response, but here we should return only one project at Services level, therefore  i wrote "byName".

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
OlegKoshelev: This method (getForUserOrFail) are used in other methods, therefore i leave it.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
OlegKoshelev: Thanks for this question! We should get either the project if there is a correct project name, or nothing (status code 404) if we transfer wrong name. I will add some fixes.

services/src/test/scala/cromwell/pipeline/service/ProjectServiceTest.scala
OlegKoshelev: I think we should get only one project at Services level whereas at level of Repository we get sequence of projects.

DruzhininPavel (approved):
Ok. Let`s rock!

gominasian (approved):
Looks fine for me

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -17,7 +17,15 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
...
+      .getProjectsByName(namePattern)
+      .flatMap(
+        projects =>
+          projects.find(_.ownerId == userId) match {
+            case Some(project) => Future.successful(project)
+            case None          => Future.failed(new ProjectNotFoundException)
+          }
+      )
 
   private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
```
myazinn: I think you must find a way to reuse this method. If we accept your changes, then we have two places that handle project visibility and they _already_ handle it differently.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
```diff
@@ -58,12 +58,14 @@ class ProjectControllerItTest
...
-        val accessToken = AccessTokenContent(stranger.userId)
+      "return a status code 404 if project doesn't exist" in {
+
+        val strangerProject = TestProjectUtils.getDummyProject(ownerId = stranger.userId)
+        val projectByNameRequest = strangerProject.name
+        val accessToken = AccessTokenContent(dummyProject.ownerId)
 
         Get("/projects?name=" + projectByNameRequest) ~> projectController.route(accessToken) ~> check {
-          status shouldBe StatusCodes.Forbidden
+          status shouldBe StatusCodes.NotFound
```
myazinn, line 76: It's a bit strange that you need to modify a test that should be unrelated to your changes.

myazinn (changes requested):

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
OlegKoshelev: It returns "NotFound" due to after changes we dont have case when method  getUserProjectByName may fail with ProjectAccessDeniedException, since we should return either project with correct name and userId or None, it seems to me.

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
OlegKoshelev: I've returned getForUserOrFail for this

myazinn: But it is the same thing as before.
You must maintain only one place that decides whether the user can see the project or not. Now we have two 
`projects => getForUserOrFail(projects.find(_.ownerId == userId), userId)`
`case Some(project) if project.ownerId == userId => Future.successful(project)`
This one place might accept several projects instead of one and then decide what it should return. It shouldn't be a problem. Remember, `Option[T]` is basically a collection of `T`s that can contain at most 1 element.

portal/src/it/scala/cromwell/pipeline/controller/ProjectControllerItTest.scala
myazinn: Previously we returned 403 if the project exists but the user doesn't have rights to see it.
How come that the ability to add projects with the same name affects it?
There are 3 cases:
1) There are no projects with such name
2) There are such projects, but the user can't see them
3) There is a project with given name that user can see

What stops us from differentiating between these three cases apart from implementation details?

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -14,13 +14,13 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
...
 
   def getUserProjectByName(namePattern: String, userId: UserId): Future[Project] =
-    projectRepository.getProjectByName(namePattern).flatMap(getForUserOrFail(_, userId))
+    projectRepository.getProjectsByName(namePattern).flatMap(getForUserOrFail(_, userId))
 
-  private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
-    project match {
+  private def getForUserOrFail(projects: Seq[Project], userId: UserId): Future[Project] =
+    projects.find(_.ownerId == userId).orElse(projects.headOption) match {
       case Some(project) if project.ownerId == userId => Future.successful(project)
```
myazinn: Much better than before, but I think we can do even better. We should be able to get rid of double check `ownerId == userId` and weird `projects.find().orElse(projects.headOption)`

OlegKoshelev: @myazinn done

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -1,29 +1,28 @@
...
-  private def getForUserOrFail(project: Option[Project], userId: UserId): Future[Project] =
-    project match {
-      case Some(project) if project.ownerId == userId => Future.successful(project)
-      case Some(_)                                    => Future.failed(new ProjectAccessDeniedException)
-      case None                                       => Future.failed(new ProjectNotFoundException)
+  private def getForUserOrFail(projects: Seq[Project], userId: UserId): Future[Project] =
+    projects.find(_.ownerId == userId) match {
+      case Some(project)             => Future.successful(project)
+      case None if projects.nonEmpty => Future.failed(new ProjectAccessDeniedException)
+      case None if projects.isEmpty  => Future.failed(new ProjectNotFoundException)
```
myazinn: Cool! Let's just empathize that the last case is the default one.
```suggestion
      case _  => Future.failed(new ProjectNotFoundException)
```
or
```suggestion
      case None  => Future.failed(new ProjectNotFoundException)
```

myazinn (approved):

GrigoriiBerezin (approved):

myazinn:
I was about to merge this PR but then I saw `chore` commit type. Please change it to a more appropriate one (I assume it would be `feat` or `fix`) and then we can merge it.
Also please add more description on how exactly you changed `getProjectByName`. What were you trying to do? How is it different from the previous implementation? Try to focus on what you are trying to do, not how you do it.

PR #173, opened 2021-06-04
EPMLSTRCMW-236

OlegKoshelev:
**DESCRIPTION**

- Right now when we increase manor/major version, child version doesn't change.

**CHANGES**

- Project.scala and VersionValue.scala have been updated
- Now if we increment the older member in the version the previous younger members must reset their values. For example we have version 1.2.3 [major.minor.revision] when increasing 
1) the major member we should get next version of 2.0.0
2) the minor member -> 1.3.0
3) the revision member -> 1.2.4

### DIFF
--- a/model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
+++ b/model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
@@ -14,6 +14,7 @@ object VersionValue extends Wrapped.Companion {
   val pattern = "^[0-9]+$"
   implicit val versionNumberFormat: Format[VersionValue] = wrapperFormat
   def increment(value: VersionValue): VersionValue = create(value.unwrap + 1)
+  def resetValue: VersionValue = create(0)
   def fromString(value: String): ValidationResult[Wrapper] =
     validateString(value) match {
       case Validated.Valid(content)  => Validated.Valid(create(content.toInt))
--- a/repositories/src/test/scala/cromwell/pipeline/datastorage/dto/PipelineVersionTest.scala
+++ b/repositories/src/test/scala/cromwell/pipeline/datastorage/dto/PipelineVersionTest.scala
@@ -1,0 +1,58 @@
+package cromwell.pipeline.datastorage.dto
+
+import cromwell.pipeline.datastorage.dto.PipelineVersion.PipelineVersionException
+import org.scalatest.{ Matchers, WordSpec }
+
+class PipelineVersionTest extends WordSpec with Matchers {
+  "PipelineVersion" when {
+
+    "apply method" should {
+      "return correct PipelineVersion" taggedAs Dto in {
+        val version = PipelineVersion("v1.2.3")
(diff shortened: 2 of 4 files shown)


DruzhininPavel (approved):
Easy peasy, but very important!
Looks good!

gominasian (approved):
Nice!

model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
```diff
@@ -14,6 +14,7 @@ object VersionValue extends Wrapped.Companion {
   val pattern = "^[0-9]+$"
   implicit val versionNumberFormat: Format[VersionValue] = wrapperFormat
   def increment(value: VersionValue): VersionValue = create(value.unwrap + 1)
+  def resetValue: VersionValue = create(0)
```
Nikolai-Isaev, line 17: Wouldn't be lazy val a better option here?

Nikolai-Isaev (approved):
Nice job, precise and accurate

model/src/main/scala/cromwell/pipeline/model/wrapper/VersionValue.scala
OlegKoshelev: @Nikolai-Isaev  I'm not sure, because I don't understand why. If you try explain this maybe I will change it.

Nikolai-Isaev: I thought that since it's value is constant-like and designed not to be changed and at the same time it can be used several times in runtime.

OlegKoshelev: I suppose that using the def will work here, since lazy variables are evaluated at runtime if I'm not mistaken.

GrigoriiBerezin (approved):
Nicely done

repositories/src/test/scala/cromwell/pipeline/datastorage/dto/PipelineVersionTest.scala
```diff
@@ -0,0 +1,58 @@
+package cromwell.pipeline.datastorage.dto
+
+import cromwell.pipeline.datastorage.dto.PipelineVersion.PipelineVersionException
+import org.scalatest.{ AsyncWordSpec, Matchers }
+
+class PipelineVersionTest extends AsyncWordSpec with Matchers {
```
myazinn:
```suggestion
class PipelineVersionTest extends WordSpec with Matchers {
```
Use `AsyncWordSpec` if you want to test Futures. Otherwise be more specific.

myazinn (approved):
Nice work!

repositories/src/test/scala/cromwell/pipeline/datastorage/dto/PipelineVersionTest.scala
OlegKoshelev: Thanks for clarification! I'll fix it.

PR #174, opened 2021-06-04
EPMLSTRCMW-250 feat: Add CORS support

AlekseiLitkovetc:
**DESCRIPTION**

In order to allow requests from the frontend side, we need to enable cross-origin requests.

**CHANGES**

- The akka-http-cors dependency was added
- The cors directive was added to the main app route
- In order to test that CORS was enabled for the main app route its definition has been moved into a separate class for convenience
- Test on enabled CORS was added
- The `ConfigJsonOps` and `ApplicationConfig` was extended
- The `ConfigJsonOpsTest` was updated
- The `.env.example` was updated

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,7 +59,7 @@ lazy val portal = project
   .settings(
     name := "Portal",
     commonSettings,
-    libraryDependencies ++= akkaDependencies ++ jsonDependencies ++ logDependencies :+ configHokon :+ akkaHttpCore,
+    libraryDependencies ++= akkaDependencies ++ jsonDependencies ++ logDependencies :+ configHokon :+ akkaHttpCore :+ akkaHttpCors,
     Defaults.itSettings,
     Seq(parallelExecution in Test := false),
     addCommandAlias("testAll", "; test ; it:test"),
--- a/portal/src/it/resources/application.conf
+++ b/portal/src/it/resources/application.conf
@@ -1,6 +1,9 @@
 webservice {
   interface = "localhost"
   port = 8080
+  cors {
+    allowedOrigins = [ "http://localhost:3000" ]
+  }
 }
 
 auth {
@@ -42,12 +45,19 @@ database {
   }
 
   mongo {
+    user = "mongoUser"
     user = ${?CMWL_TEST_MONGO_USER}
+    password = "password"
     password = ${?CMWL_TEST_MONGO_PASSWORD}
+    host = "localhost"
     host = ${?CMWL_TEST_MONGO_HOST}
+    port = "27017"
     port = ${?CMWL_TEST_MONGO_PORT}
+    authenticationDatabase = "admin"
     authenticationDatabase = ${?CMWL_TEST_MONGO_AUTH}
+    database = "mongo"
     database = ${?CMWL_TEST_MONGO_DB}
+    collection = "configurations"
     collection = ${?CMWL_TEST_MONGO_COLLECTION}
   }
 }
--- a/portal/src/main/scala/cromwell/pipeline/CromwellPipelineRoute.scala
+++ b/portal/src/main/scala/cromwell/pipeline/CromwellPipelineRoute.scala
@@ -1,0 +1,37 @@
+package cromwell.pipeline
+
+import akka.http.scaladsl.model.headers.HttpOrigin
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import ch.megard.akka.http.cors.javadsl.model.HttpOriginMatcher
+import ch.megard.akka.http.cors.scaladsl.settings.CorsSettings
+import cromwell.pipeline.controller.ControllerModule
+import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.utils.ApplicationConfig
+
+object CromwellPipelineRoute {
+  type SecuredRoute = AccessTokenContent => Route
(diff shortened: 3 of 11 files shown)


portal/src/it/resources/application.conf
```diff
@@ -42,12 +42,23 @@ database {
   }
 
   mongo {
+    user = "mongoUser"
```
DruzhininPavel, line 48: I don`t understand, why do you use static parameters in this place? We have all parameters in .env file.

DruzhininPavel (changes requested):
This is a great job!
But only one thing. application.conf seems to me to be restored back.

portal/src/it/resources/application.conf
AlekseiLitkovetc: @DruzhininPavel, thank you for your review.
I'm happy to see that you follow the changes closely.

But I would like to highlight that this request was in the draft state and it had WIP (work in progress) in the title.
So, it hasn't asked for a review at that moment, I was working on it to check some things.

Moving back to your suggestion, yes, we have `.env` file with params but it's not provided for our IT tests when we run them.
These static params using just in the IT's `application.conf`, not in the main one.
I would prefer to leave it as is.

portal/src/it/resources/application.conf
```diff
@@ -42,12 +42,23 @@ database {
...
+    database = "mongo"
     database = ${?CMWL_TEST_MONGO_DB}
+    collection = "configurations"
     collection = ${?CMWL_TEST_MONGO_COLLECTION}
   }
 }
+
+akka-http-cors {
+  allowed-origins = [ "http://localhost:3000" ]
+}
```
myazinn: Shouldn't we put it along with other settings for webservice like `host` and `port`?

portal/src/main/scala/cromwell/pipeline/CromwellPipelineRoute.scala
```diff
@@ -0,0 +1,23 @@
+package cromwell.pipeline
+
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Route
+import cromwell.pipeline.CromwellPipelineApp.routeCombiner
```
myazinn: We should move `routeCombiner` here too. Otherwise we have a bit complex structure (`CromwellPipelineRoute` and `CromwellPipelineApp` depend on each other).

build.sbt
```diff
@@ -59,7 +59,7 @@ lazy val portal = project
   .settings(
     name := "Portal",
     commonSettings,
-    libraryDependencies ++= akkaDependencies ++ jsonDependencies ++ logDependencies :+ configHokon :+ akkaHttpCore,
+    libraryDependencies ++= akkaDependencies ++ jsonDependencies ++ logDependencies :+ configHokon :+ akkaHttpCore :+ akkaHttpCors,
```
myazinn, line 62: [minor]
I think we should add `akkaHttpCors` and `akkaHttpCore` to `akkaDependencies`. Or at least combine `akkaHttpCore` into `akkaHttpCors` somehow. Right now it exposes to many dependency details.

portal/src/it/resources/application.conf
DruzhininPavel: Oops. Sorry. I didn't saw that it was draft. I'll be better in future.

DruzhininPavel (dismissed):

portal/src/it/resources/application.conf
myazinn: Also, please use camelCase to match the current config style.

myazinn (changes requested):
Nice work! I have a few minor comments though

DruzhininPavel (approved):

portal/src/it/resources/application.conf
AlekseiLitkovetc: It's okay,
Thank you!

portal/src/it/resources/application.conf
AlekseiLitkovetc: It would be reasonable to do so
Done

portal/src/main/scala/cromwell/pipeline/CromwellPipelineRoute.scala
AlekseiLitkovetc: You are correct
Thank you

build.sbt
AlekseiLitkovetc: I don't think that this is a good idea to group it this way. By doing this we will bring excess dependencies to modules.
The `akkaDependencies` and `akkaHttpCore` are used in several places whereas akkaHttpCors should be used only in the `portal` module.
I propose to keep the current implementation.

myazinn: Got it. I agree, we should keep it as is.

portal/src/it/resources/application.conf
```diff
@@ -1,6 +1,9 @@
 webservice {
   interface = "localhost"
   port = 8080
+  akkaHttpCors {
```
myazinn: [minor]
I think we can use just `cors`.

myazinn (approved):
Awesome!

portal/src/it/resources/application.conf
AlekseiLitkovetc: Let's use `cors`

Nikolai-Isaev (approved):
Great job!

GrigoriiBerezin (approved):

gominasian (approved):
So nice!

PR #177, opened 2021-06-10
EPMLSTRCMW-251 Feat: Generate exception in LiquidBaseUtils

OlegKoshelev:
DESCRIPTION

In the liquibase utils in case of error we're just logging message. There is no good in this. Please make it return the error, and pass this error right to the Main class. Also consider failing pipeline start if migration is failed.

CHANGES

- LiquibaseUtils.scala, PipelineDatebaseEngine.scala and CromwellPipelineApp.scala have been updated
-  Now updateSchema method may return a liquibaseUpdateSchemaException which may be thown to main class and the application stops as a result.

### DIFF
--- a/datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
+++ b/datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
@@ -1,5 +1,6 @@
 package cromwell.pipeline.database
 
+import cromwell.pipeline.database.LiquibaseExceptions.LiquibaseUpdateSchemaException
 import liquibase.database.DatabaseFactory
 import liquibase.database.jvm.JdbcConnection
 import liquibase.resource.ClassLoaderResourceAccessor
@@ -15,17 +16,26 @@ object LiquibaseUtils {
   private val DefaultContexts = new Contexts()
   private val DefaultLabelExpression = new LabelExpression()
 
-  def updateSchema(changeLogResourcePath: String)(jdbcConnection: Connection): Unit = {
+  def updateSchema(
+    changeLogResourcePath: String
+  )(jdbcConnection: Connection): Try[Unit] = {
     val liquibaseConnection = new JdbcConnection(jdbcConnection)
+
     Try {
       val database = DatabaseFactory.getInstance.findCorrectDatabaseImplementation(liquibaseConnection)
       val liquibase = new Liquibase(changeLogResourcePath, new ClassLoaderResourceAccessor(), database)
 
       liquibase.update(DefaultContexts, DefaultLabelExpression)
     } match {
-      case Failure(exception) => log.error("An error during the schema update via liquibase.", exception)
-      case Success(_)         => log.error("Successfully updated db schema via liquibase!")
+      case Failure(exception) =>
+        liquibaseConnection.close()
+        val liquibaseUpdateSchemaException = LiquibaseUpdateSchemaException(exception)
+        log.error(liquibaseUpdateSchemaException.message, liquibaseUpdateSchemaException.cause)
+        Failure(liquibaseUpdateSchemaException)
+      case Success(_) =>
+        liquibaseConnection.close()
+        log.info("Successfully updated db schema via liquibase!")
+        Success(())
     }
-    liquibaseConnection.close()
   }
 }
--- a/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
+++ b/portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
@@ -27,25 +27,40 @@ object CromwellPipelineApp extends App {
     .result()
     .withFallback(RejectionHandler.default)
 
-  Try {
-    val components = new ApplicationComponents()
+  def initializeApplicationComponents: Try[ApplicationComponents] = Try {
+    new ApplicationComponents()
+  }
 
-    import components.applicationConfig
-    import components.applicationConfig.webServiceConfig
+  def updateDb(components: ApplicationComponents): Try[Unit] = {
     import components.datastorageModule._
-
     pipelineDatabaseEngine.updateSchema()
+  }
 
-    val route = new CromwellPipelineRoute(components.applicationConfig, components.controllerModule).route
+  def startHttpServer(components: ApplicationComponents): Try[Unit] =
+    Try {
+      import components.applicationConfig
+      import components.applicationConfig.webServiceConfig
+      val route = new CromwellPipelineRoute(applicationConfig, components.controllerModule).route
+      log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
+      log.info(ConfigJsonOps.configToJsonString(applicationConfig))
+      Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
+    }
 
-    log.info(s"Server online at http://${webServiceConfig.interface}:${webServiceConfig.port}/")
-    log.info(ConfigJsonOps.configToJsonString(applicationConfig))
-    Http().newServerAt(webServiceConfig.interface, webServiceConfig.port).bindFlow(route)
-  } match {
-    case Success(_) =>
-      log.info(s"Application is running!")
-    case Failure(e) =>
+  val app = for {
+
+    applicationComponents <- initializeApplicationComponents
+
+    _ <- updateDb(applicationComponents)
+
+    _ <- startHttpServer(applicationComponents)
+
+  } yield ()
+
+  app match {
+    case Success(_) => log.info(s"Application is running!")
+
+    case Failure(exception) =>
+      log.error(s"Application cannot be started:", exception)
       system.terminate()
-      log.error(s"Application cannot be started:", e)
   }
 }
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
@@ -3,7 +3,7 @@ package cromwell.pipeline.service
 import cromwell.pipeline.datastorage.dao.repository.ProjectRepository
 import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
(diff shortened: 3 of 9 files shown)


repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
```diff
@@ -60,6 +60,7 @@ trait Profile {
       new StringColumnExtensionMethods[String](c.asInstanceOf[Rep[String]])
 
     private def iso[A, B](map: A => B, comap: B => A) = new Isomorphism(map, comap)
+
```
DruzhininPavel: You should remove this :)

DruzhininPavel (approved):
LGTM

repositories/src/main/scala/cromwell/pipeline/datastorage/DatastorageModule.scala
OlegKoshelev: @DruzhininPavel done

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
```diff
@@ -15,17 +16,30 @@ object LiquibaseUtils {
...
+      case Success(_) =>
+        liquibaseConnection.close()
+        log.info("Successfully updated db schema via liquibase!")
+        Success(())
     }
-    liquibaseConnection.close()
   }
 }
+
+object Exception {
```
Nikolai-Isaev: That naming may be not very informative. 
Btw, what is the reason to wrap that exception into an object. Couldn't we just put it inside LiquiBaseUtils?

Nikolai-Isaev (approved):
Nice job

Nikolai-Isaev:
Commit message is a little bit confusing, as if these changes are supposed to make errors)

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
OlegKoshelev: I've seen a similar approach with other exceptions in current project, so I used it.

OlegKoshelev:
> Commit message is a little bit confusing, as if these changes are supposed to make errors)

It seems to me that the commit messages should be short, if you want to get more information, you can always look at the description of the problem.

myazinn:
> > Commit message is a little bit confusing, as if these changes are supposed to make errors)
> 
> It seems to me that the commit messages should be short, if you want to get more information, you can always look at the description of the problem.

@OlegKoshelev 
Yes, should be short, but it doesn't mean that it can't be informative. Right now commit message says that your commit will create an error in `LiquidBaseUtils`. Why would we merge something that intentionally creates errors in our project? It doesn't make sense.
Although we can get more information from description, it takes much more time in this case. Since this commit message doesn't have descriptive body, developer would have to find original PR in Github. In other words, it will take minutes instead of seconds.
Lastly, please fix your commit type. According to our documentation,
```
- `chore`: updating build tasks, package manager configs, etc; no production code change
```
Your PR definitely affects production code and no builds tasks / package managers are involved. I don't understand why you chose `chore` as a commit type.

@Nikolai-Isaev is totally right, we should pay a lot of attention to our commit messages in order to spare time of other developers in the future.

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
myazinn: We might have overlooked it accidentally.
That's not a good thing, please change this name to something more descriptive. And if you see uninformative name somewhere else - feel free to rename it too.

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
```diff
@@ -15,17 +16,30 @@ object LiquibaseUtils {
...
       val database = DatabaseFactory.getInstance.findCorrectDatabaseImplementation(liquibaseConnection)
       val liquibase = new Liquibase(changeLogResourcePath, new ClassLoaderResourceAccessor(), database)
 
       liquibase.update(DefaultContexts, DefaultLabelExpression)
     } match {
-      case Failure(exception) => log.error("An error during the schema update via liquibase.", exception)
-      case Success(_)         => log.error("Successfully updated db schema via liquibase!")
+      case Failure(exception) =>
+        liquibaseConnection.close()
+        val liquibaseUpdateSchemaException = LiquibaseUpdateSchemaException()
```
myazinn: Since this exception is going to be propagated further, it's worth adding additional information on what went wrong. You can add it as a `cause` field (for example, take a look at available constructors of `RuntimeException`)

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
```diff
@@ -15,17 +16,30 @@ object LiquibaseUtils {
...
+        log.error(liquibaseUpdateSchemaException.message, exception)
+        Failure(liquibaseUpdateSchemaException)
+      case Success(_) =>
+        liquibaseConnection.close()
+        log.info("Successfully updated db schema via liquibase!")
```
myazinn, line 37: Nice catch with `log.error` -> `log.info`!

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -35,7 +35,6 @@ object CromwellPipelineApp extends App {
     import components.datastorageModule._
 
     pipelineDatabaseEngine.updateSchema()
```
myazinn, line 36: We should not ignore result of this `update` anymore. If it returns `Failure` then we should not start an application.

myazinn (changes requested):

OlegKoshelev:
> > > Commit message is a little bit confusing, as if these changes are supposed to make errors)
> > 
> > 
> > It seems to me that the commit messages should be short, if you want to get more information, you can always look at the description of the problem.
> 
> @OlegKoshelev
> Yes, should be short, but it doesn't mean that it can't be informative. Right now commit message says that your commit will create an error in `LiquidBaseUtils`. Why would we merge something that intentionally creates errors in our project? It doesn't make sense.
> Although we can get more information from description, it takes much more time in this case. Since this commit message doesn't have descriptive body, developer would have to find original PR in Github. In other words, it will take minutes instead of seconds.
> Lastly, please fix your commit type. According to our documentation,
> 
> ```
> - `chore`: updating build tasks, package manager configs, etc; no production code change
> ```
> 
> Your PR definitely affects production code and no builds tasks / package managers are involved. I don't understand why you chose `chore` as a commit type.
> 
> @Nikolai-Isaev is totally right, we should pay a lot of attention to our commit messages in order to spare time of other developers in the future.

Ok. I'll fix it.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
OlegKoshelev: @myazinn It seems to me that the exception should be caught on line 46 and the application will be terminated. Or is it necessary to implement it somehow differently?

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
OlegKoshelev: Thanks, I'll glance

datasource/src/main/scala/cromwell/pipeline/database/LiquibaseUtils.scala
OlegKoshelev: Ok, I'll fix it.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
myazinn: There is no exception that will be thrown. Try to add invalid migration and see what happens.
Probably you need to read something about the difference between `try {} catch {}` and `Try()` in Scala.

gominasian (approved):
LGTM

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -60,9 +60,3 @@ class ProjectService(projectRepository: ProjectRepository, projectVersioning: Pr
       projectRepository.updateProjectVersion(project.copy(version = version))
     }
 }
-
-object Exceptions {
```
myazinn, line 64: Thanks for noticing it!
There is a more standard way to handle it. Instead of creating `ProjectServiceExceptions`, we can put these exceptions into `ProjectService` companion object.
```
object ProjectService {
  object Exceptions {
    ???
  }
}
```
That way, if we want to use such exception then it will look almost identical to your implementation. I.e. `ProjectService.Exceptions` instead of `ProjectServiceExceptions`. Why bother changing it? Because in this case it would be much easier to find them.
You probably should apply the same approach to `Liquibase` exceptions.

portal/src/main/scala/cromwell/pipeline/CromwellPipelineApp.scala
```diff
@@ -34,8 +34,7 @@ object CromwellPipelineApp extends App {
     import components.applicationConfig.webServiceConfig
     import components.datastorageModule._
 
-    pipelineDatabaseEngine.updateSchema()
-
+    pipelineDatabaseEngine.updateSchema().get
```
myazinn: Avoid using `.get` on data types like `Option`, `Try`, `Either`, etc. Scala provides you more delicate ways to work with them like `map` / `flatMap` and `for-comprehension`.
In your case, `.get` makes your code exception-driven which is a bad practice even in Java

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
OlegKoshelev: Thanks for clarification! I'll change it.

myazinn (approved):

GrigoriiBerezin (approved):

Dmitrii-Shevchenko (approved):

PR #179, opened 2021-06-14
EPMLSTRCMW-227 chore: add PR name validation

myazinn:


DruzhininPavel (approved):
Looks good to me. :)

.github/workflows/PR_info_check.yml
```diff
@@ -6,12 +6,33 @@ on:
...
 
 jobs:
   check-commit-message:
     name: Check Commit Message
     runs-on: ubuntu-latest
     steps:
+      - name: Check PR JIRA reference
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: '^(\[WIP\] )?((EPMLSTRCMW-\d{1,4})|(IS-00)) .*$'
```
Nikolai-Isaev: Shouldn't we use a commit type here also?

Nikolai-Isaev (approved):
LGTM

.github/workflows/PR_info_check.yml
myazinn: Currently we don't have such requirement anywhere.
We can discuss it, update documentation and then add such check. But I don't think it's necessary because, as the name implies, it's a _commit_ type :) PR's name doesn't have to strictly match commit message and PR might even have several commits with different types.

OlegKoshelev (approved):

GrigoriiBerezin (approved):

.github/workflows/PR_info_check.yml
GrigoriiBerezin: maybe
```suggestion
          pattern: '^(\[WIP\] )?((EPMLSTRCMW-\d+)|(IS-00)) .*$'
```
to not care at all about task numbers?

Dmitrii-Shevchenko (approved):

.github/workflows/PR_info_check.yml
myazinn: Done

gominasian (approved):
Nice!

PR #185, opened 2021-07-02
EPMLSTRCMW-265 refactor: Replacing UserRepository with UserService

OlegKoshelev:
**DESCRIPTION**

Use `UserService` in `AuthService` instead of `UserRepository`

**CHANGES**
- AuthService.scala, ServiceModule.scala, UserService.scala, AuthServiceTest.scala and UserServiceTest.scala have been updated.
- It allows to avoid a break of encapsulation and future problems during refactoring.

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/ServiceModule.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ServiceModule.scala
@@ -15,7 +15,7 @@ class ServiceModule(
 )(
   implicit executionContext: ExecutionContext
 ) {
-  lazy val authService: AuthService = AuthService(datastorageModule.userRepository, authModule.authUtils)
+  lazy val authService: AuthService = AuthService(userService, authModule.authUtils)
   lazy val userService: UserService = UserService(datastorageModule.userRepository)
   lazy val projectVersioning: GitLabProjectVersioning = new GitLabProjectVersioning(httpClient, config)
   lazy val projectService: ProjectService = ProjectService(datastorageModule.projectRepository, projectVersioning)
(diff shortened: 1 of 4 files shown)


DruzhininPavel (approved):
LGTM

Nikolai-Isaev (approved):
Lgtm

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -50,3 +55,8 @@ class UserService(userRepository: UserRepository)(implicit executionContext: Exe
       }
     } else Future.failed(new RuntimeException("new password incorrectly duplicated"))
 }
+
+object UserService {
+  def apply(userRepository: UserRepository)(implicit executionContext: ExecutionContext): UserService =
+    new UserService(userRepository)(executionContext)
+}
```
myazinn: This change is ok by itself, but it will make code style in `UserService` inconsistent with other services. We should either change all services to have an `apply` method, or don't change this one. Since we already have a task EPMLSTRCMW-241 where we create `apply` method for all services, I suggest you to just revert this change

gominasian (approved):

services/src/main/scala/cromwell/pipeline/service/UserService.scala
OlegKoshelev: @myazinn done

services/src/main/scala/cromwell/pipeline/service/ServiceModule.scala
```diff
@@ -16,7 +16,7 @@ class ServiceModule(
   implicit executionContext: ExecutionContext
 ) {
   lazy val authService: AuthService =
-    new AuthService(datastorageModule.userRepository, authModule.authUtils)
+    new AuthService(new UserService(datastorageModule.userRepository), authModule.authUtils)
```
myazinn: Sorry for not noticing it earlier, but why do we create another instance of `UserService` here? Can't we use `userService` defined on the line below?

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/ServiceModule.scala
OlegKoshelev: Thanks! I did not notice

OlegKoshelev: @myazinn done

myazinn (approved):

Dmitrii-Shevchenko (approved):

PR #189, opened 2021-07-05
EPMLSTRCMW-264 fix : users search by email

OlegKoshelev:
**DESCRIPTION**

Users search by email returns sensitive data (URL returns all info about user, including his password hash and salt)

**CHANGES**

- Tests have been fixed
- Now  URL returns base info about user (id, email, firstName, lastName, active)


postman/Users.postman_collection.json
```diff
@@ -1,289 +1,289 @@
...
+              "key": "Content-Type",
+              "value": "application/json"
+            },
+            {
+              "key": "Content-Length",
+              "value": "292"
+            }
+          ],
+          "cookie": [],
+          "body": "[\n    {\n        \"userId\": \"26774154-8946-4764-b8ab-37dfeccd3453\",\n        \"email\": \"example@mail.ru\",\n        \"firstName\": \"Ivan\",\n        \"lastName\": \"Ivanov\",\n        \"active\": true\n    }\n]"
```
Nikolai-Isaev: It is really hard to find the difference here :) Is it only that line?

Nikolai-Isaev (approved):
Lgtm

postman/Users.postman_collection.json
OlegKoshelev: @Nikolai-Isaev yes, I have changed it for correct display of "Find user by email example" in postman.

controllers/src/main/scala/cromwell/pipeline/controller/UserController.scala
```diff
@@ -15,7 +16,7 @@ class UserController(userService: UserService) {
   private val getUser: Route = get {
     parameter('email.as[String]) { email =>
       onComplete(userService.getUsersByEmail(email)) {
-        case Success(r) => complete(r)
+        case Success(usersWithCredentials) => complete(usersWithCredentials.map(User.fromUserWithCredentials))
```
DruzhininPavel: I think we can add `apply(user: UserWithCredentials): User` method to the object User and some implicit method like implicit `def fromUserWithCredentials(user: UserWithCredentials): User = User(user)`
Or we can create some converting from Seq[UserWithCredentials] to Seq[User] with implicits
I think it would be better for the project and coders.
But it`s up tou you

DruzhininPavel (approved):
Look`s good. But I have one question.

gominasian (approved):
LGTM

controllers/src/main/scala/cromwell/pipeline/controller/UserController.scala
OlegKoshelev: Thanks for comment. 
1) I think apply method may lead to worse understanding of code. Also previous implementation contained fromUser method.
2) During mapping smth we should clearly specify method otherwise compiler will not be able to understand  what type we should get.
3) For creating converting from Seq[UserWithCredentials] to Seq[User] we also should specify type clearly Seq[User] to compilator will be able to eliminate type mismatch.

@myazinn what do you think?

myazinn: I don't see why we will experience compile issues.
`User.fromUserWithCredentials(userWithCredentials)` can be converted to `User.apply(userWithCredentials)` which can be further converted to `User(userWithCredentials)`. At which step do you think the problem will arise?
In this particular case we can go even further and replace it with underscore `usersWithCredentials.map(User(_))`. It will work just fine.

myazinn:
1) Why does your commit type is `refactor`? You are fixing a bug and even you commit message header starts with `fix`.
2) I'm surprised that this PR changes so many classes. Suggested in ticket fix should have effectively affected only some places that directly work with the user password. Are you sure that we really need `UserWithCredentials` everywhere?

OlegKoshelev:
> 1. Why does your commit type is `refactor`? You are fixing a bug and even you commit message header starts with `fix`.
> 2. I'm surprised that this PR changes so many classes. Suggested in ticket fix should have effectively affected only some places that directly work with the user password. Are you sure that we really need `UserWithCredentials` everywhere?

1. I was not sure if this was a bug since "Find user by email example" in postman  contained all information about users. But I'll change it to "fix".
2.  I think I should take it back to User and UserNoCredentials. I have made so since condition of this task required it.

controllers/src/main/scala/cromwell/pipeline/controller/UserController.scala
OlegKoshelev: > I don't see why we will experience compile issues.
> `User.fromUserWithCredentials(userWithCredentials)` can be converted to `User.apply(userWithCredentials)` which can be further converted to `User(userWithCredentials)`. At which step do you think the problem will arise?
> In this particular case we can go even further and replace it with underscore `usersWithCredentials.map(User(_))`. It will work just fine.

Ok. I'll add apply method. I just wrote the map method requires clearly specify converting even if we use implicit methods.

myazinn:
1) Have you took a look at ticket type in JIRA? :)
2) Are you planning to get rid of `password` field in `User`?

OlegKoshelev:
> 1. Have you took a look at ticket type in JIRA? :)
> 2. Are you planning to get rid of `password` field in `User`?

1) Ok, I understood.
2) Do you mean both passwordHash and passwordSalt?

myazinn:
Yes, I mean both

OlegKoshelev:
> 1. Have you took a look at ticket type in JIRA? :)
> 2. Are you planning to get rid of `password` field in `User`?

I just map User => UserNoCredentials since it doesn't contain password fields.

myazinn:
Sorry, but I disagree with this decision. It will fix only one symptom of the real problem and we will definitely face something similar again.
The problem is that we use `User` class everywhere and it contains passwordHash and passwordSalt which ideally should be used _only_ in authentication service. Other places should not know about it nor be allowed to use it. We should a version with and without the password. We already have them under different names, that's why fix in jira suggest to just rename them to `UserWithCredentials` and `User` respectively. And then we can carefully use `UserWithCredentials` in some specific cases where we really need to. Why rename them at all and not just replace usage in some cases? Because most of the time we don't care about the password and people would tend to use `User` class instead of `UserNoCredentials` because it sounds better and more intuitive. 
We've already discussed this solution during our daily meetings, but we can discuss it again if required 
cc: @ilyaonishenko

DruzhininPavel (approved):
A lot of work has been done.
But maybe you shouldn't shift the work of the service layer to the database? I would leave the full User in the database, and on the service layer I would remove the credentilas from him using something like user.map (UserWOCredentials) and give it to everyone. In the opposite direction, approximately the same, we take the User from the database and update the necessary fields, then write back.
What are you thinking about this?

OlegKoshelev:
> A lot of work has been done.
> But maybe you shouldn't shift the work of the service layer to the database? I would leave the full User in the database, and on the service layer I would remove the credentilas from him using something like user.map (UserWOCredentials) and give it to everyone. In the opposite direction, approximately the same, we take the User from the database and update the necessary fields, then write back.
> What are you thinking about this?

1.Could you describe problems which can emerge in more detail since I don't understand why presented solution isn't fit? 
2. From my side may tell full user are stored in DB and you may put full user with all fields to DB. I tried to avoid the getting of full user due to probability to give it to controllers level. I think using current approach allows to avoid it.

DruzhininPavel:
> > A lot of work has been done.
> > But maybe you shouldn't shift the work of the service layer to the database? I would leave the full User in the database, and on the service layer I would remove the credentilas from him using something like user.map (UserWOCredentials) and give it to everyone. In the opposite direction, approximately the same, we take the User from the database and update the necessary fields, then write back.
> > What are you thinking about this?
> 
> 1.Could you describe problems which can emerge in more detail since I don't understand why presented solution isn't fit?
> 2. From my side may tell full user are stored in DB and you may put full user with all fields to DB. I tried to avoid the getting of full user due to probability to give it to controllers level. I think using current approach allows to avoid it.

1. Yes, of course I can describe. The main problem is the over-complicated approach and, as a consequence, a more difficult understanding of what is happening. We write not only for ourselves but also for those who will come after us. And of course it will be easier for them to understand if there is one entity in the database with which we work, and all the business logic for processing, transforming and transferring this entity will pass on the service layer. This is at least good form, as well as fast and high-quality code support.

2. I absolutely agree, but according to the accepted rules, the database should not process entities in any way, ideally it should only receive and save it as it is, or give it completely as it is stored. In this case, it is much easier for us to migrate and do not have to resort to some tricky methods to collect entities from several into one or from one to many. I believe that simplicity is the key to the health of the project.

OlegKoshelev:
> > > A lot of work has been done.
> > > But maybe you shouldn't shift the work of the service layer to the database? I would leave the full User in the database, and on the service layer I would remove the credentilas from him using something like user.map (UserWOCredentials) and give it to everyone. In the opposite direction, approximately the same, we take the User from the database and update the necessary fields, then write back.
> > > What are you thinking about this?
> > 
> > 
> > 1.Could you describe problems which can emerge in more detail since I don't understand why presented solution isn't fit?
> > 2. From my side may tell full user are stored in DB and you may put full user with all fields to DB. I tried to avoid the getting of full user due to probability to give it to controllers level. I think using current approach allows to avoid it.
> 
> 1. Yes, of course I can describe. The main problem is the over-complicated approach and, as a consequence, a more difficult understanding of what is happening. We write not only for ourselves but also for those who will come after us. And of course it will be easier for them to understand if there is one entity in the database with which we work, and all the business logic for processing, transforming and transferring this entity will pass on the service layer. This is at least good form, as well as fast and high-quality code support.
> 2. I absolutely agree, but according to the accepted rules, the database should not process entities in any way, ideally it should only receive and save it as it is, or give it completely as it is stored. In this case, it is much easier for us to migrate and do not have to resort to some tricky methods to collect entities from several into one or from one to many. I believe that simplicity is the key to the health of the project.

1. I focused on safety because student with small experience may not think about it when he is trying to get User since it may not matter for him. Thus it may leads to leakage of sensitive data. In addition the understanding of all User entities it's not big deal, it just requires to dig a bit  deeper. Achieving security requires the sacrifice of convenience. Moreover credentials are used only in one place during authentication whereas getting user without credentials are used ubiquitous.
2.  I think it will not leads to problem with migration since in current implementation we have entity of full user with all fields because it is used for saving users. Furthermore getting entities at database level will be faster since in this case we get only fields are needed for us instead of getting other object from full user.

DruzhininPavel:
> * I focused on safety because student with small experience may not think about it when he is trying to get User since it may not matter for him. Thus it may leads to leakage of sensitive data. In addition the understanding of all User entities it's not big deal, it just requires to dig a bit  deeper. Achieving security requires the sacrifice of convenience. Moreover credentials are used only in one place during authentication whereas getting user without credentials are used ubiquitous.
> * I think it will not leads to problem with migration since in current implementation we have entity of full user with all fields because it is used for saving users. Furthermore getting entities at database level will be faster since in this case we get only fields are needed for us instead of getting other object from full user.

Ok. I have no questions.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -62,22 +62,23 @@ object UserService {
...
-                case _ => Future.failed(new RuntimeException("user password differs from entered"))
-              }
-            case None => Future.failed(new RuntimeException("user with this id doesn't exist"))
-          }
-        } else Future.failed(new RuntimeException("new password incorrectly duplicated"))
-
+        request.newPassword match {
+          case request.repeatPassword =>
+            userRepository.getUserById(userId).flatMap {
+              case Some(userWithCredentials)
```
DruzhininPavel: I think. we don't need to write this value as `userWithCredentials`, I think that` user` should be enough.

DruzhininPavel (approved):
Looks better than the previous one. But only one comment.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -11,7 +11,7 @@ import scala.util.Random
 
 trait UserService {
 
-  def getUsersByEmail(emailPattern: String): Future[Seq[User]]
+  def getUsersByEmail(emailPattern: String): Future[Seq[UserNoCredentials]]
 
   def getUserByEmail(email: UserEmail): Future[Option[User]]
```
Nikolai-Isaev: Shouldn't the same approach be implemented to that method as well?

controllers/src/test/scala/cromwell/pipeline/controller/UserControllerTest.scala
```diff
@@ -36,16 +36,17 @@ class UserControllerTest
       "return the sequence of users" taggedAs Controller in {
         val usersByEmailRequest: UserEmail = UserEmail("someDomain@mail.com", Enable.Unsafe)
         val dummyUser: User = TestUserUtils.getDummyUser()
+        val dummyUserNoCredentials = UserNoCredentials.fromUser(dummyUser)
         val userId = dummyUser.userId
-        val uEmailRespSeq: Seq[User] = Seq(dummyUser)
+        val uEmailRespSeq: Seq[UserNoCredentials] = Seq(dummyUserNoCredentials)
```
myazinn: Can you please rename current `User` to `UserWithCredentials` and `UserNoCredentials` to `User`? Since user without creds should be our default choice now, it's better to give it a convenient name

services/src/main/scala/cromwell/pipeline/service/UserService.scala
myazinn: We need this method for authentication, so it must return a full user. However, I suggest renaming it to `getUserWithCredsByEmail` or something like that to explicitly show that it shouldn't be used arbitrarily.

myazinn (changes requested):

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -62,22 +62,23 @@ object UserService {
         request: PasswordUpdateRequest,
```
Nikolai-Isaev: Is this just a refactoring or are these changes connected to the current task? Anyway, i would appreciate a small explanation on what is the main reason for these changes.

gominasian (approved):
Nice job!

controllers/src/main/scala/cromwell/pipeline/controller/UserController.scala
DruzhininPavel: Done.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
OlegKoshelev: I just a bit rewrote the implementation of this method since  the previous version was a little confusing.

services/src/main/scala/cromwell/pipeline/service/UserService.scala
OlegKoshelev: I did not change it because it better reflects reality.

controllers/src/test/scala/cromwell/pipeline/controller/UserControllerTest.scala
OlegKoshelev: @myazinn done

services/src/main/scala/cromwell/pipeline/service/UserService.scala
DruzhininPavel: Ok. As you wish.

controllers/src/test/scala/cromwell/pipeline/controller/UserWithCredentialsControllerTest.scala
```diff
@@ -17,7 +17,7 @@ import org.scalatestplus.mockito.MockitoSugar
 
 import scala.concurrent.Future
 
-class UserControllerTest
+class UserWithCredentialsControllerTest
```
myazinn: Can we please keep `UserControllerTest` ? Same for `UserController`, `UserServiceTest` etc. As we already discussed, these classes are not supposed to work with `UserWithCredentials`, so why call them that way?

controllers/src/test/scala/cromwell/pipeline/controller/UserWithCredentialsControllerTest.scala
```diff
@@ -35,9 +35,10 @@ class UserControllerTest
 
       "return the sequence of users" taggedAs Controller in {
         val usersByEmailRequest: UserEmail = UserEmail("someDomain@mail.com", Enable.Unsafe)
-        val dummyUser: User = TestUserUtils.getDummyUser()
+        val dummyUser: UserWithCredentials = TestUserUtils.getDummyUser()
```
myazinn: Now once you renamed `User` to `UserWithCredentials` it is clear that we have a lot of places that use full `UserWithCredentials` although they shouldn't do it. <br>For example, here we definitely don't need full `UserWithCredentials` and need only simple `User`. So why bother creating it? I suggest you to tweak `TestUserUtils.getDummyUser()` to return `User` (the one without creds) and create separate `TestUserUtils.getDummyUserWithCredentials()` for some places where we really need a full user.<br>
tip: after you introduce `TestUserUtils.getDummyUserWithCredentials()`, you can re-use in `TestUserUtils.getDummyUser()`

myazinn (changes requested):

controllers/src/test/scala/cromwell/pipeline/controller/UserWithCredentialsControllerTest.scala
OlegKoshelev: @myazinn done

controllers/src/test/scala/cromwell/pipeline/controller/UserWithCredentialsControllerTest.scala
OlegKoshelev: @myazinn done

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
```diff
@@ -22,7 +22,7 @@ class ProjectControllerTest extends AsyncWordSpec with Matchers with ScalatestRo
 
   private val projectService = mock[ProjectService]
   private val projectController = new ProjectController(projectService)
-  private val stranger = TestUserUtils.getDummyUser()
+  private val stranger = TestUserUtils.getDummyUserWithCredentials()
```
myazinn: Why do we need user with credentials here?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/User.scala
```diff
@@ -2,25 +2,25 @@ package cromwell.pipeline.datastorage.dto
...
-  active: Boolean = true
+  active: Boolean
 )
 
 object User {
-  implicit lazy val userFormat: OFormat[User] = Json.format[User]
-}
-
-final case class ProfilePicture(value: Array[Byte]) extends MappedTo[Array[Byte]]
+  implicit val UserNoCredentialsFormat: OFormat[User] =
```
myazinn:
```suggestion
  implicit val userFormat: OFormat[User] =
```

myazinn (changes requested):
Awesome work! Only two minor annoying things left

controllers/src/test/scala/cromwell/pipeline/controller/ProjectControllerTest.scala
OlegKoshelev: @myazinn I fixed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/User.scala
OlegKoshelev: @myazinn done

myazinn (approved):
Thanks for this amazing work!

Dmitrii-Shevchenko (approved):

PR #195, opened 2021-07-27
EPMLSTRCMW-267 refactor: Unnest UserService.updatePassword and make it more readable

myazinn:


Dmitrii-Shevchenko (approved):
Looks exciting

DruzhininPavel (approved):
Super !, The Password wrapper is finally in use!

services/src/main/scala/cromwell/pipeline/service/UserService.scala
```diff
@@ -61,23 +61,40 @@ object UserService {
...
+            Future.unit
+          } else {
+            Future.failed(new RuntimeException("new password incorrectly duplicated"))
+          }
+
+        def checkUserPassword(user: User): Future[Unit] =
+          if (user.passwordHash == calculatePasswordHash(request.currentPassword, user.passwordSalt)) {
+            Future.unit
+          } else {
+            Future.failed(new RuntimeException("new password incorrectly duplicated"))
```
Nikolai-Isaev: Shouldn't this exception message be about wrong current password or something like that?

myazinn: Ooops. Nice catch!

myazinn: I'm surprised that all tests pass :) Fixed this error and added a few tests to make sure it won't happen in the future

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/UserRepositoryTestImp.scala
```diff
@@ -39,7 +39,8 @@ class UserRepositoryTestImp extends UserRepository {
 
   def updateUser(updatedUser: User): Future[Int] = update(updatedUser)
 
-  def updatePassword(updatedUser: User): Future[Int] = update(updatedUser)
+  def updatePassword(userId: UserId, hash: String, salt: String): Future[Int] =
+    users.get(userId).fold(Future.successful(0))(u => update(u.copy(passwordHash = hash, passwordSalt = salt)))
```
Nikolai-Isaev, line 43: If i got it right, we check whether there is a user under provided userId, and if there is, we check the same condition again in the update method. May be in that case it will be better to delete private update method instead of reusing it?

Nikolai-Isaev (approved):
Great!

repositories/src/test/scala/cromwell/pipeline/datastorage/dao/repository/UserRepositoryTestImp.scala
myazinn: It bothers me a bit as well, but if I remove `update` method and inline it's code in `updateUser` and `updatePassword` methods, then we will have to copy-paste some code. IMO double check with fairly readable code is better than code duplication.

Nikolai-Isaev: Got it, thanks!

gominasian (approved):
Nice!

iliaonishchenko (approved):
Looks good to me 👍

PR #196, opened 2021-07-28
[WIP] EPMLSTRCMW-243 Create test implementation of service classes

Nikolai-Isaev:
feat: Create test implementation of service classes


DruzhininPavel (commented):
I think that you shold transfer all methods from Classes to Objects in apply method. and to use traits.

PR #199, opened 2021-08-03
EPMLSTRCMW-273 docs: add documentation on Wrapped

myazinn:


docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+      user.gender,
+      user.name,
+      user.department,
+      user.avatar
+    )
+  }
+}
+```
+I hope now you see we have added this trait to our project.
+However, you probably noticed that our `Wrapped` contains not only `toString`, but also a bunch of other methods.<br>
```
Nikolai-Isaev: Can't notice that actually :) Seems there have to be these methods

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+    println(s"Sending [$content] to $email")
+  }
+}
+
+class LawsInfoService {
+  def canHaveRussianPassword(age: Int): Boolean = {
+    if (!Utils.isValidAge(age)) throw new RuntimeException("age is invalid")
+    age >= 14
+  }
+}
```
DruzhininPavel: I think there is a mistake here. method 'canHaveRussianPassword'. as I know, should not containt validation of age.

DruzhininPavel (approved):

docs/what-is-wrapped.md
myazinn: Funny typo :) Fixed

docs/what-is-wrapped.md
myazinn: Ooops. It should be `real Wrapped` instead of `our Wrapped`. Fixed, thanks!

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+  // hashcode takes into account wrapper class information and its content. This is done to obey equals / hashcode laws
+  override def hashCode: Int = this.getClass.hashCode + value.hashCode()
+}
+
+object Wrapped {
+  trait Companion {
+    type Type
+    type Wrapper <: Wrapped[Type]
+    
+    // we don't have to define such alias, but lest do it in order to show what this Either means
```
Nikolai-Isaev: Typo in "lest", i believe

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+}
+
+class LawsInfoService {
+  def canHaveRussianPassport(age: Age): Boolean =
+    age.value >= 14
+}
+
+class UserRepository {
+  def get(id: UserId): User = ???
+    // oops, it doesn't work anymore. We will deal with it later
```
Nikolai-Isaev: Don't know if that is even an issue, but that "oops, it doesn't work anymore" holds from the previous example on 735 line. Thought, that I scrolled up accidentally. Maybe something like "it still doesn't work, but wait a little and we will handle it with the most brilliant fashion" would be less confusing :)

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+## Conclusion
+Now, go to the real project and take a look at `Wrapped` and `Wrapped.Companion`. You should be able to understand the purpose of every line in these traits.<br>
+Still not convinced that it is not an overengineered mess? That's ok! Try to implement the same functionality by yourself.
+Odds are you either end up with something very similar to current implementation and understand why do we need every line (which is great) or create a better solution (even better!).
+If latter is the case, feel free to open a pull request and fix current implementation.
+
+
+
+## Useful information
+We are not the first people who faced such problems. A lot of people noticed it way earlier and created libraries that can done such things automatically.<br>
```
Nikolai-Isaev: Typo in "done"

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+    type ValidationResult[A] = Either[Error, A]
+ 
+    protected def validate(value: Type): ValidationResult[Type]
+    protected def create(value: Type): Wrapper
+
+    final def from(value: Type): ValidationResult[Wrapper] = validate(value).map(create)
+  
+    final def uglyUnsafeCreation(value: Type): Wrapper = from(value) match {
+      case Right(wrapper) => wrapper
+      case Left(error) => throw new RuntimeException(s"Oh no! You promised me that it's ok to unsafely wrap value but I got an error! $error")
```
Nikolai-Isaev: That one is still vague. How it helps to create unsafely if it still returns only a valid result. And yes, you promised that it's ok to unsafely wrap value but we got an error! Why?

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1007 @@
...
+  age: Age,
+  gender: Gender,
+  password: Password,
+  email: Email,
+  department: Department,
+  avatar: Avatar
+)
+
+class UserRepository {
+  // let's forget about this unsafe place. Usually it's done by DB libraries that won't mess up
```
Nikolai-Isaev: Why is this place "unsafe"

Nikolai-Isaev (approved):
That is a fabulous article that must be published, no jokes. Great job, please write more.

docs/what-is-wrapped.md
myazinn: I have added comment to clarify it

docs/what-is-wrapped.md
myazinn: Thanks, fixed

docs/what-is-wrapped.md
myazinn: Agree. Changed to `still doesn't work. We will fix it soon`

docs/what-is-wrapped.md
myazinn: Fixed, thanks

docs/what-is-wrapped.md
myazinn: "unsafe" here doesn't mean we allow creation from invalid data. Thanks for noticing it, I've explained it in detail in updated document. Please take a look.

```
Actually we can just leave it as is and force people to do something like this `Email.from("some@email.com").getOrElse(throw new RuntimeException("invalid email)`.
But again, it will lead to code duplication, and it's better to have one common place for such things with standard error message to make our life easier.<br>
Therefore, we need to give people a way to create wrapped instances directly.
Let's do it by introducing new method for unsafe creation. However, we will do this method so ugly, that it would be impossible to use it without being noticed.
Yes, it would be unsafe and an obvious flaw, but hopefully it would be a controlled place that everyone is aware of.<br>
_Note_. Word "unsafe" here doesn't mean we allow creation from invalid data. It's still strictly forbidden. "Unsafe" means that our factory method will return `Wrapped` directly instead of `Either[Error, Wrapped]` so that people don't have to get rid of this `Either` manually.
I.e. we are replacing compile-time checks with runtime checks.
```

gominasian (approved):
I think you should write a book. It's the greatest explanation I've ever read

Dmitrii-Shevchenko (approved):

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1014 @@
+# What are `Wrapped[T]` and `Wrapped.Companion` ? 
+
+I assume you've already started your work on Crowmwell Pipeline and found `Wrapped[T]` trait that contains a bunch of methods.
+Also, you probably noticed its companion object with a mindblowing `Companion` trait. Take a look at it if not.
+It contains a lot of things that at first glance might look overengineered. Don't worry, you are not the first one who got confused otherwise this page wouldn't exist.<br>
```
iliaonishchenko: `<br>` is from HTML syntax, not markdown 😄

docs/what-is-wrapped.md
```diff
@@ -0,0 +1,1014 @@
...
+  private val dummyEmailRegexp: String = "(\\w|\\d)+@(\\w|\\d)+\\.(\\w|\\d)+"
+  def fromString(value: String): Either[String, Email] =
+    if (value.matches(dummyEmailRegexp)) Right(new Email(value)) else Left(s"email must match regexp $dummyEmailRegexp")
+    
+}
+//val email: Email = new Email("not an email") // won't compile anymore
+//val email: Email = Email.fromString("not an email") // won't compile anymore
+val email: Either[String, Email] = Email.fromString("not an email")
+```
+Notice that we don't use `case class` anymore. This is because `case class`es by their nature are open data structures that can be easily created. We don't want it anymore.
```
iliaonishchenko, line 691: It's not a remark, but an addition. It's possible to make case classes not so open and with removed `copy` method.
For example
```scala
sealed abstract case class Email private (value: String)

object Email {
  def fromString(emailStr: String): Option[Email] = {
    Some(new Email(emailStr) {} )
  }
}
```

And in this case hashcode and equals I assume still exist and work correctly.

iliaonishchenko (approved):
Wow! It's a great thing!

docs/what-is-wrapped.md
myazinn: Worked fine though :) changed `<br>` to `\`

docs/what-is-wrapped.md
myazinn: Thanks for reminding it, you are right. I've added your example to document.

PR #201, opened 2021-08-04
EPMLSTRCMW-258 Feat: Change endpoints in ConfigurationController to RESTful

DruzhininPavel:
**Description:**
Changed endpoints in ConfigurationController to be more RESTful

**Changes:**
ProjectConfigurationController - changed endpoint methods
ProjectConfigurationControllerTest - changed 5 tests
ProjectServiceController - changed addConfiguration method
ProjectServiceControllerTest - changed 3 tests
ProjectConfiguration - added CongugurationAdditionRequest
postman cillection - fixed about new endpoint

### DIFF
--- a/controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
+++ b/controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
@@ -1,7 +1,5 @@
 package cromwell.pipeline.controller
 
-import java.nio.file.Paths
-
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.testkit.ScalatestRouteTest
 import cromwell.pipeline.datastorage.dao.utils.{ TestProjectUtils, TestUserUtils }
@@ -13,6 +11,7 @@ import org.mockito.Mockito.when
 import org.scalatest.{ AsyncWordSpec, Matchers }
 import org.scalatestplus.mockito.MockitoSugar
 
+import java.nio.file.Paths
 import scala.concurrent.Future
 
 class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers with ScalatestRouteTest with MockitoSugar {
@@ -21,14 +20,21 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
 
   "ProjectConfigurationController" when {
     val accessToken = AccessTokenContent(TestUserUtils.getDummyUserId)
+    val projectId = TestProjectUtils.getDummyProjectId
     val configuration = ProjectConfiguration(
       ProjectConfigurationId.randomId,
-      TestProjectUtils.getDummyProjectId,
+      projectId,
       active = true,
       List(
         ProjectFileConfiguration(Paths.get("/home/file"), List(FileParameter("nodeName", StringTyped(Some("hello")))))
       ),
       ProjectConfigurationVersion.defaultVersion
+    )
+    val configurationAdditionRequest = ProjectConfigurationAdditionRequest(
+      id = configuration.id,
+      active = configuration.active,
+      projectFileConfigurations = configuration.projectFileConfigurations,
+      version = configuration.version
     )
 
     "update configuration" should {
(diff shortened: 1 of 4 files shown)


services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -30,25 +28,38 @@ object ProjectConfigurationService {
...
     new ProjectConfigurationService {
 
-      def addConfiguration(projectConfiguration: ProjectConfiguration, userId: UserId): Future[Unit] = {
+      def addConfiguration(
+        projectId: ProjectId,
+        projectConfigurationRequest: ProjectConfigurationAdditionRequest,
+        userId: UserId
+      ): Future[Unit] = {
+
+        val mewProjectConfiguration = ProjectConfiguration(
```
Nikolai-Isaev: "mew" - That's what I call a real cats effect :)

Nikolai-Isaev (approved):
Looks good but tricky to me

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
DruzhininPavel: Ooops. I`ll fix it.

myazinn:
I don't think we should change these routes, at least not the way you did it.
`ProjectConfiguration` has its own `id` and is determined by it, not by `projectId`.
If I call `/configurations/some-random-uuid` then I would expect the endpoint to return `ProjectConfiguration` where `id = some-random-uuid`, not `projectId = some-random-uuid`. In our case `configurations/projectId=???` is basically a search query. Moreover, if at some point we decide to allow users to fetch configurations by their id directly, then we will have no way of doing it.
@ilyaonishenko what do you think?

DruzhininPavel:
> I don't think we should change these routes, at least not the way you did it.
> `ProjectConfiguration` has its own `id` and is determined by it, not by `projectId`.
> If I call `/configurations/some-random-uuid` then I would expect the endpoint to return `ProjectConfiguration` where `id = some-random-uuid`, not `projectId = some-random-uuid`. In our case `configurations/projectId=???` is basically a search query. Moreover, if at some point we decide to allow users to fetch configurations by their id directly, then we will have no way of doing it.
> @ilyaonishenko what do you think?

I think we shouldn't think about the endpoints in more detail. Since the entities of the ProjectConfiguration type are nested in the Project, maybe we should put them in the appropriate paths by analogy with something like /project/{{som-project-uuid}}/configurations/{{som-configutaion-uuid}}  this would lead to an understanding of the logic of the endpoints and the correct their construction.

Nikolai-Isaev (approved):

gominasian (approved):
LGTM

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
```diff
@@ -1,21 +1,19 @@
...
+import cromwell.pipeline.datastorage.dto._
 import cromwell.pipeline.model.wrapper.UserId
 
 import java.nio.file.Path
 import scala.concurrent.{ ExecutionContext, Future }
 
 trait ProjectConfigurationService {
 
-  def addConfiguration(projectConfiguration: ProjectConfiguration, userId: UserId): Future[Unit]
+  def addConfiguration(
```
myazinn: I just realized that it's wrong to change signatures of `Service` methods when we change API scheme. This method should still accept whole `ProjectConfiguration`, now it's up to `Controller` layer to assemble it.

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
```diff
@@ -21,50 +20,63 @@ class ProjectConfigurationControllerTest extends AsyncWordSpec with Matchers wit
...
 
     "update configuration" should {
       val error = new RuntimeException("Something went wrong")
 
       "return success for update configuration" in {
-        when(configurationService.addConfiguration(configuration, accessToken.userId)).thenReturn(Future.unit)
-        Put("/configurations", configuration) ~> configurationController.route(accessToken) ~> check {
+        when(configurationService.addConfiguration(projectId, configurationAdditionRequest, accessToken.userId))
+          .thenReturn(Future.unit)
+        Put(s"/projects/${projectId.value}/configurations", configurationAdditionRequest) ~> configurationController
```
myazinn, line 45:
```suggestion
        Put(s"/projects/$projectId/configurations", configurationAdditionRequest) ~> configurationController
```
Same for other places

myazinn (changes requested):

controllers/src/test/scala/cromwell/pipeline/controller/ProjectConfigurationControllerTest.scala
DruzhininPavel: If we change this, then we get somthing like this "/project/Project("{PROJECT_UUID}")/configuration"

myazinn: Oh. Ok, nevermind then

services/src/main/scala/cromwell/pipeline/service/ProjectConfigurationService.scala
DruzhininPavel: I thout, that controller didn`t create new entities. 
But ok, I can change this.

myazinn: It should not contain businesses logic, but it is allowed to trivially map data from one form to another.
The fact that `projectId` comes from different place than other configuration info is purely caused by our changes of API scheme, no business logic involved. So now it's up to controller layer to make inputs match `Service` layer expectations.

DruzhininPavel: Done

myazinn (approved):

Dmitrii-Shevchenko (approved):

PR #203, opened 2021-08-04
EPMLSTRCMW-271 Fix RunService addRun method to check ProjectId before create

DruzhininPavel:
**Description:**
addRun method can add Run object into any project
we must fix it.

**Changes:**
RunService - added check into addRun method
RunServiceTest - fixed test
ServiceModule - fixed creating runService

### DIFF
--- a/services/src/main/scala/cromwell/pipeline/service/RunService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/RunService.scala
@@ -21,13 +21,12 @@ trait RunService {
 
 object RunService {
 
-  def apply(runRepository: RunRepository)(
+  def apply(runRepository: RunRepository, projectService: ProjectService)(
     implicit executionContext: ExecutionContext
   ): RunService =
     new RunService {
 
       def addRun(runCreateRequest: RunCreateRequest, userId: UserId): Future[RunId] = {
-
         val newRun = Run(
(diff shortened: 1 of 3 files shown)


services/src/main/scala/cromwell/pipeline/service/RunService.scala
```diff
@@ -37,7 +36,7 @@ object RunService {
           userId = userId,
           results = runCreateRequest.results
         )
-        runRepository.addRun(newRun)
+        projectService.getUserProjectById(runCreateRequest.projectId, userId).flatMap(_ => runRepository.addRun(newRun))
```
Nikolai-Isaev, line 39: If I'm right `addRun` will now cause two more exceptions from `getUserProjectById`. Shouldn't exceptions be handled more informatively  in `RunController`'s `addRun` from now on?

DruzhininPavel: Yep. we can add more exceptions, but I think we should do this on new task.

Nikolai-Isaev (approved):
Lgtm

myazinn (approved):

gominasian (approved):
Nice!

Dmitrii-Shevchenko (approved):

PR #216, opened 2021-08-26
EPMLSTRCMW-287 Create test implementations for ProjectFile and ProjectVersioning services

Nikolai-Isaev:
test: Add ProjectFile and ProjectVersioning test implementations

### DIFF
--- a/services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
@@ -1,0 +1,60 @@
+package cromwell.pipeline.service
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.model.wrapper.UserId
+
+import java.nio.file.Path
+import scala.concurrent.Future
+
+class ProjectFileServiceTestImpl(projectFiles: Seq[ProjectFile], testMode: TestMode) extends ProjectFileService {
+
+  override def validateFile(fileContent: ProjectFileContent): Future[Either[ValidationError, Unit]] =
+    testMode match {
+      case WithException(exc: ValidationError) => Future.successful(Left(exc))
+      case WithException(exc)                  => Future.failed(exc)
+      case _                                   => Future.successful(Right(()))
+    }
+
+  override def uploadFile(
+    projectId: ProjectId,
+    projectFile: ProjectFile,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[Either[VersioningException, Unit]] =
+    testMode match {
+      case WithException(exc: VersioningException) => Future.successful(Left(exc))
+      case WithException(exc)                      => Future.failed(exc)
+      case _                                       => Future.successful(Right(()))
+    }
+
+  override def getFile(
+    projectId: ProjectId,
+    path: Path,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[ProjectFile] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _                  => Future.successful(projectFiles.find(_.path == path).get)
+    }
+
+  override def getFiles(
+    projectId: ProjectId,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[List[ProjectFile]] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _                  => Future.successful(projectFiles.toList)
+    }
+
+}
+
+object ProjectFileServiceTestImpl {
+
+  def apply(projectFiles: ProjectFile*): ProjectFileServiceTestImpl =
+    new ProjectFileServiceTestImpl(projectFiles = projectFiles, testMode = Success)
+
+  def withException(exception: Throwable): ProjectFileServiceTestImpl =
+    new ProjectFileServiceTestImpl(projectFiles = Seq.empty, testMode = WithException(exception))
+
+}
--- a/services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
@@ -1,0 +1,75 @@
+package cromwell.pipeline.service
+import cromwell.pipeline.datastorage.dto._
+
+import java.nio.file.Path
+import scala.concurrent.Future
+
+class ProjectVersioningTestImpl(
+  projects: List[Project],
+  pipelineVersions: List[PipelineVersion],
+  projectFiles: List[ProjectFile],
+  commits: List[Commit],
+  testMode: TestMode
+) extends ProjectVersioning[VersioningException] {
+
+  override def createRepository(localProject: LocalProject): AsyncResult[Project] =
+    returnValOrExc(projects.head)
+
+  override def updateFile(
+    project: Project,
+    projectFile: ProjectFile,
+    version: Option[PipelineVersion]
+  ): AsyncResult[PipelineVersion] =
+    returnValOrExc(pipelineVersions.head)
+
+  override def updateFiles(
+    project: Project,
+    projectFiles: ProjectFiles,
+    version: Option[PipelineVersion]
+  ): AsyncResult[PipelineVersion] =
+    returnValOrExc(pipelineVersions.head)
+
+  override def getFile(project: Project, path: Path, version: Option[PipelineVersion]): AsyncResult[ProjectFile] =
+    returnValOrExc(projectFiles.find(_.path == path).get)
+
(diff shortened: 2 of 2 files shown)


services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def getFile(
+    projectId: ProjectId,
+    path: Path,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[ProjectFile] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _                  => Future.successful(projectFiles.head)
```
DruzhininPavel: Maybe we want to take the file at least by name? 
Then we can write something like:
```suggestion
      case _                  => Future.successful(projectFiles.filter(_.path == path).take(1))
```

services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
```diff
@@ -0,0 +1,91 @@
...
+      case WithException(exc: VersioningException) => Future.successful(Left(exc))
+      case WithException(exc)                      => Future.failed(exc)
+      case _                                       => Future.successful(Right(pipelineVersions.head))
+    }
+
+  override def getFile(project: Project, path: Path, version: Option[PipelineVersion]): AsyncResult[ProjectFile] =
+    testMode match {
+      case WithException(exc: VersioningException) => Future.successful(Left(exc))
+      case WithException(exc)                      => Future.failed(exc)
+      case _                                       => Future.successful(Right(projectFiles.head))
```
DruzhininPavel: In this place to. If we have more then one file we never get that what we whant, only the first.

DruzhininPavel (changes requested):
I have only one small change for you.

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: I thought that we will create a new testImpl for every testcase. So to that testImpl we will pass objects which we want to return. And if we want to receive one specific file we will pass just that single file. We just don't need to pass a whole bunch to choose from. Please, correct me if I'm wrong.

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
```diff
@@ -0,0 +1,60 @@
...
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.model.wrapper.UserId
+
+import java.nio.file.Path
+import scala.concurrent.Future
+
+class ProjectFileServiceTestImpl(projectFiles: Seq[ProjectFile], testMode: TestMode) extends ProjectFileService {
+
+  override def validateFile(fileContent: ProjectFileContent): Future[Either[ValidationError, Unit]] =
+    testMode match {
```
gominasian, line 11: I don't really get the point of TestMode. If you could explain it to me it would be great. Everything else seems right to me

gominasian (approved):

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: It is a marker, which tells testImpl to return a result needed for current testcase. It allows to simulate two types of behavior - success or failure - inside one class. 
`val projectFileService = ProjectFileServiceTestImpl(dummyFile)` - if we want a method to be successfull or
`val projectFileService = ProjectFileServiceTestImpl.withException(someException)` - if we want a method with an exception.
In both cases these values are instances of the same `ProjectFileServiceTestImpl`. `TestMode` is needed to tell the test implementation which type of behavior it must choose.

Other way could be creating two different test Implementations, for example `ProjectFileServiceFailure` - it's overridden methods always fail, and `ProjectFileServiceSuccessfull` - always works without errors.

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
DruzhininPavel: I just think this will be more expected behavior from a test implementation. It's just that if we decide to write a test where there will be several elements, then we cannot be sure that the one we need will be returned to us.

Nikolai-Isaev: Fixed, thanks

services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
Nikolai-Isaev: Fixed

DruzhininPavel: Super

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
DruzhininPavel: Ok.

DruzhininPavel (approved):

services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
```diff
@@ -0,0 +1,100 @@
...
+      case WithException(exc: VersioningException) => Future.successful(Left(exc))
+      case WithException(exc)                      => Future.failed(exc)
+      case _                                       => Future.successful(Right(pipelineVersions))
+    }
+
+  override def getFileVersions(project: Project, path: Path): AsyncResult[List[PipelineVersion]] =
+    testMode match {
+      case WithException(exc: VersioningException) => Future.successful(Left(exc))
+      case WithException(exc)                      => Future.failed(exc)
+      case _                                       => Future.successful(Right(pipelineVersions))
```
myazinn: In other services it was tolerable, but in this service we have a LOT of similar code. Can you please de-duplicate it?

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
```diff
@@ -0,0 +1,60 @@
...
+      case WithException(exc: ValidationError) => Future.successful(Left(exc))
+      case WithException(exc)                  => Future.failed(exc)
+      case _                                   => Future.successful(Right(Unit))
```
myazinn:
```suggestion
      case _                                   => Future.successful(Right(()))
```
`Unit` here is not the same thing as `()`. `Unit` is a type that has a companion object called `Unit` (the one that you return here) and the instance of `Unit` can be created with `()`. Here it compiles because Scala treats methods that return `Unit` specially, you can write 
```suggestion
      case _                                   => Future.successful(Right("hello"))
```
It will work, be you shouldn't do it :)

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  override def getFile(
+    projectId: ProjectId,
+    path: Path,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[ProjectFile] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _                  => Future.successful(projectFiles.filter(_.path == path).head)
```
myazinn:
```suggestion
      case _                  => Future.successful(projectFiles.find(_.path == path).get)
```

myazinn (changes requested):

services/src/test/scala/cromwell/pipeline/service/ProjectVersioningTestImpl.scala
Nikolai-Isaev: Fixed

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: Shame on me. Fixed

services/src/test/scala/cromwell/pipeline/service/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: Done

GrigoriiBerezin (approved):

myazinn (approved):
nice

PR #224, opened 2021-09-21
EPMLSTRCMW-290 Replace mocks with test implementations in service tests

Nikolai-Isaev:
test: Replace mocks with impls in service

### DIFF
--- a/services/src/test/scala/cromwell/pipeline/service/impls/WomToolTestImpl.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/impls/WomToolTestImpl.scala
@@ -1,0 +1,50 @@
+package cromwell.pipeline.service.impls
+
+import cats.data.NonEmptyList
+import cromwell.pipeline.datastorage.dto.FileParameter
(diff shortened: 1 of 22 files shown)


services/src/test/scala/cromwell/pipeline/service/impls/WomToolTestImpl.scala
```diff
@@ -0,0 +1,50 @@
...
+      override def validate(content: String): Either[NonEmptyList[String], WomBundle] =
+        getMessages
+
+      override def stringInputs(content: String): Either[NonEmptyList[String], String] =
+        getMessages
+
+      override def inputsToList(content: String): Either[NonEmptyList[String], List[FileParameter]] =
+        getMessages
+
+      private def getMessages[T]: Either[NonEmptyList[String], T] =
```
DruzhininPavel, line 46: it might be worth adding parentheses to a method like `getMessages [T](): `? This helps us to understand that this is a method.

DruzhininPavel (changes requested):
One small change from me.

PR #225, opened 2021-09-30
EPMLSTRCMW-292 Refactor custom exceptions for RunService

DruzhininPavel:


services/src/main/scala/cromwell/pipeline/service/RunService.scala
```diff
@@ -38,36 +48,69 @@ object RunService {
...
+                  cmwlWorkflowId = request.cmwlWorkflowId
+                )
               )
-            )
-          case None => Future.failed(new RuntimeException("run with this id doesn't exist"))
+              .recoverWith { case _ => internalError("update run") }
+          case None => Future.failed(NotFound())
+        }
+
+      private def internalError(action: String): Future[Nothing] =
```
Nikolai-Isaev, line 105: That method name does not look like an action. Maybe "getInternalError" or smth like that would be better

Nikolai-Isaev (approved):
Nice

services/src/main/scala/cromwell/pipeline/service/RunService.scala
```diff
@@ -38,36 +48,69 @@ object RunService {
           userId = userId,
           results = runCreateRequest.results
         )
-        projectService.getUserProjectById(projectId, userId).flatMap(_ => runRepository.addRun(newRun))
+        projectService
+          .getUserProjectById(projectId, userId)
+          .recoverWith {
+            case e: ProjectServiceException => serviceErrorMapper(e)
+          }
+          .flatMap(_ => runRepository.addRun(newRun))
```
Nikolai-Isaev, line 56: Can `transform` be used here?

services/src/main/scala/cromwell/pipeline/service/RunService.scala
DruzhininPavel: This name reflects the results you get and is familiar. We do not pass the result outside the class boundaries. In addition, in other classes, it is also named the same, I think it is not worth changing traditions :)

services/src/main/scala/cromwell/pipeline/service/RunService.scala
DruzhininPavel: Do you have a specific example ?

gominasian (approved):
Looks good to me

myazinn (approved):

GrigoriiBerezin (approved):
Nice

PR #226, opened 2021-09-30
EPMLSTRCMW-290 Replace mocks with test implementations in service tests

Nikolai-Isaev:
test: Replace mocks with impls in service

### DIFF
--- a/auth/src/main/scala/cromwell/pipeline/auth/AuthUtils.scala
+++ b/auth/src/main/scala/cromwell/pipeline/auth/AuthUtils.scala
@@ -5,57 +5,73 @@ import cromwell.pipeline.utils.AuthConfig
 import pdi.jwt.{ Jwt, JwtClaim }
 import play.api.libs.json.Json
 
-class AuthUtils(authConfig: AuthConfig) {
+trait AuthUtils {
 
   def getAuthResponse(
     accessTokenContent: AccessTokenContent,
     refreshTokenContent: RefreshTokenContent,
     currentTimestamp: Long
-  ): Option[AuthResponse] = {
+  ): Option[AuthResponse]
 
-    import authConfig._
+  def getOptJwtClaims(refreshToken: String): Option[JwtClaim]
 
-    def buildAuthResponse: AuthResponse = {
-      val updatedAccessTokenLifetime = refreshTokenContent.optRestOfUserSession.map { restOfUserSession =>
-        Seq(restOfUserSession, expirationTimeInSeconds.accessToken).min
-      }.getOrElse(expirationTimeInSeconds.accessToken)
+}
 
-      val updatedRefreshTokenLifetime = refreshTokenContent.optRestOfUserSession.map { restOfUserSession =>
-        Seq(restOfUserSession, expirationTimeInSeconds.refreshToken).min
-      }.getOrElse(expirationTimeInSeconds.refreshToken)
+object AuthUtils {
 
-      val updatedRefreshTokenContent = refreshTokenContent.optRestOfUserSession.map { restOfUserSession =>
-        val updatedRestOfUserSession = restOfUserSession - updatedRefreshTokenLifetime
-        refreshTokenContent.copy(optRestOfUserSession = Some(updatedRestOfUserSession))
-      }.getOrElse {
-        val restOfUserSession = expirationTimeInSeconds.userSession - expirationTimeInSeconds.refreshToken
-        refreshTokenContent.copy(optRestOfUserSession = Some(restOfUserSession))
+  def apply(authConfig: AuthConfig): AuthUtils =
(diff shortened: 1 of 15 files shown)


auth/src/main/scala/cromwell/pipeline/auth/AuthUtils.scala
```diff
@@ -5,57 +5,73 @@ import cromwell.pipeline.utils.AuthConfig
...
-  }
+          AuthResponse(newAccessToken, newRefreshToken, currentTimestamp + updatedAccessTokenLifetime)
+        }
 
+        refreshTokenContent.optRestOfUserSession match {
+          case optRestOfUserSession if optRestOfUserSession.exists(_ > 0) || optRestOfUserSession.isEmpty =>
+            Some(buildAuthResponse)
+          case _ =>
+            None
+        }
```
DruzhininPavel, line 61: I think this could be changed for more understanding. But it's up to you to decide.
```suggestion
        refreshTokenContent.optRestOfUserSession match {
          case optRestOfUserSession if optRestOfUserSession.exists(_ <= 0) => None
          case _                                                                                                   => Some(buildAuthResponse)
        }
```

DruzhininPavel (approved):
Great job, I just have one comment.

gominasian (approved):

auth/src/main/scala/cromwell/pipeline/auth/AuthUtils.scala
Nikolai-Isaev: I prefer not to change code which is not directly involved in the task

services/src/test/scala/cromwell/pipeline/service/impls/WomToolTestImpl.scala
```diff
@@ -0,0 +1,50 @@
...
+        getMessages
+
+      override def stringInputs(content: String): Either[NonEmptyList[String], String] =
+        getMessages
+
+      override def inputsToList(content: String): Either[NonEmptyList[String], List[FileParameter]] =
+        getMessages
+
+      private def getMessages[T]: Either[NonEmptyList[String], T] =
+        Left(NonEmptyList.fromList(errorMessages).get)
```
myazinn:
```suggestion
        Left(NonEmptyList.fromListUnsafe(errorMessages))
```

services/src/test/scala/cromwell/pipeline/service/RunServiceTest.scala
```diff
@@ -41,40 +33,47 @@ class RunServiceTest extends AsyncWordSpec with Matchers with MockitoSugar {
...
+
+    "addRun" should {
       "returns run id" taggedAs Service in {
         val runCreateRequest = RunCreateRequest(projectVersion = run.projectVersion, results = run.results)
-        when(projectService.getUserProjectById(projectId, userId)).thenReturn(Future.successful(project))
-        when(runRepository.addRun(any[Run])).thenReturn(Future.successful(runId))
 
         runService.addRun(runCreateRequest, projectId, userId).map {
-          _ shouldBe runId
+          _ shouldBe a[RunId]
```
myazinn: Obviously it will be `a[RunId]`, that's what the type system will enforce even without tests. Why did you need to change this assertion?

myazinn (changes requested):

services/src/test/scala/cromwell/pipeline/service/RunServiceTest.scala
Nikolai-Isaev: In that case `repository.addRun` returns a random run Id, generated inside `runService.addRun`.
Can't think of the decent way of how to test it with current repository testImpl.

services/src/test/scala/cromwell/pipeline/service/impls/WomToolTestImpl.scala
Nikolai-Isaev: Fixed

services/src/test/scala/cromwell/pipeline/service/RunServiceTest.scala
myazinn: In this case, there are two ways to fix is.
1) Make your assertion more explicit 
```suggestion
          _ => succeed
```
2) Change the test order and modify this one. You should move tests for `getRunByIdAndUser` method on top and then add a test for `addRun` method. This test will add a run, then fetch it with `getRunByIdAndUser` and compare that this is the same run as we expected.

Nikolai-Isaev: Great ideas, thanks! Fixed

myazinn (approved):

GrigoriiBerezin (approved):

PR #229, opened 2021-10-25
EPMLSTRCMW-299 Implement search system

Nikolai-Isaev:
feat: Implement search system

### DIFF
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ControllerModule.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ControllerModule.scala
@@ -18,4 +18,6 @@ class ControllerModule(serviceModule: ServiceModule, authConfig: AuthConfig)(
   lazy val projectFileController: ProjectFileController = new ProjectFileController(serviceModule.projectFileService)
   lazy val configurationController = new ProjectConfigurationController(serviceModule.configurationService)
   lazy val runController: RunController = new RunController(serviceModule.runService)
+  lazy val projectSearchController: ProjectSearchController =
+    new ProjectSearchController(serviceModule.projectSearchService)
 }
--- a/controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
+++ b/controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
@@ -1,0 +1,36 @@
+package cromwell.pipeline.controller
+
+import akka.http.scaladsl.model.StatusCodes
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.{ ExceptionHandler, Route }
+import cromwell.pipeline.controller.ProjectSearchController.projectSearchExceptionHandler
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.service.ProjectSearchService
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+
+class ProjectSearchController(projectSearchService: ProjectSearchService) {
+
+  private def searchProjects(implicit accessToken: AccessTokenContent): Route = post {
+    pathEndOrSingleSlash {
+      entity(as[ProjectSearchRequest]) { request =>
+        complete(projectSearchService.searchProjects(request, accessToken.userId))
+      }
+    }
+
+  }
+
+  val route: AccessTokenContent => Route = implicit accessToken =>
+    handleExceptions(projectSearchExceptionHandler) {
+      pathPrefix("projects" / "search") {
+        searchProjects
+      }
+    }
+}
+
+object ProjectSearchController {
+
+  val projectSearchExceptionHandler: ExceptionHandler = ExceptionHandler {
+    case e => complete(StatusCodes.InternalServerError, e.getMessage)
+  }
+}
--- a/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
+++ b/repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
@@ -1,0 +1,94 @@
+package cromwell.pipeline.datastorage.dto
+
+import cromwell.pipeline.utils.json.AdtJsonFormatter._
+import play.api.libs.functional.syntax.{ toFunctionalBuilderOps, toInvariantFunctorOps, unlift }
+import play.api.libs.json.Format.GenericFormat
+import play.api.libs.json.{ Json, __, _ }
+
+sealed trait ProjectSearchFilter
+case object All extends ProjectSearchFilter
+final case class ByName(mode: NameSearchMode, value: String) extends ProjectSearchFilter
+final case class ByConfig(mode: ContentSearchMode, value: Boolean) extends ProjectSearchFilter
+final case class ByFiles(mode: ContentSearchMode, value: Boolean) extends ProjectSearchFilter
+final case class Or(left: ProjectSearchFilter, right: ProjectSearchFilter) extends ProjectSearchFilter
+final case class And(left: ProjectSearchFilter, right: ProjectSearchFilter) extends ProjectSearchFilter
+
+object ProjectSearchFilter {
+
+  implicit val projectSearchFilterFormat: OFormat[ProjectSearchFilter] = {
+
+    implicit val allFormat: OFormat[All.type] = objectFormat(All)
+    implicit val byNameFormat: OFormat[ByName] = Json.format
+    implicit val byConfigFormat: OFormat[ByConfig] = Json.format
+    implicit val byFilesFormat: OFormat[ByFiles] = Json.format
+    implicit val orFormat: OFormat[Or] = getComplexFilterFormat(Or, Or.unapply)
+    implicit val andFormat: OFormat[And] = getComplexFilterFormat(And, And.unapply)
+
+    adtFormat("type")(
+      adtCase[All.type]("all"),
+      adtCase[ByName]("name"),
+      adtCase[ByFiles]("files"),
+      adtCase[ByConfig]("configurations"),
+      adtCase[Or]("or"),
+      adtCase[And]("and")
+    )
+  }
+  private def getComplexFilterFormat[T](
+    apply: (ProjectSearchFilter, ProjectSearchFilter) => T,
+    unapply: T => Option[(ProjectSearchFilter, ProjectSearchFilter)]
+  ): OFormat[T] = OFormat(
+    ((__ \ "left").lazyRead(projectSearchFilterFormat) ~
+      (__ \ "right").lazyRead(projectSearchFilterFormat))(apply),
+    ((__ \ "left").lazyWrite(projectSearchFilterFormat) ~
+      (__ \ "right").lazyWrite(projectSearchFilterFormat))(unlift(unapply))
+  )
+}
+
+sealed trait NameSearchMode
+case object FullMatch extends NameSearchMode
+case object RegexpMatch extends NameSearchMode
+
+object NameSearchMode {
+  implicit lazy val nameSearchModeFormat: Format[NameSearchMode] = {
+    implicitly[Format[String]].inmap(NameSearchMode.fromString, NameSearchMode.toString)
+  }
+
+  def fromString(s: String): NameSearchMode = s match {
+    case "full_match"   => FullMatch
+    case "regexp_match" => RegexpMatch
+  }
+
+  def toString(mode: NameSearchMode): String = mode match {
+    case FullMatch   => "full_match"
+    case RegexpMatch => "regexp_match"
+  }
+}
+
+sealed trait ContentSearchMode
+case object Exists extends ContentSearchMode
+
+object ContentSearchMode {
+  implicit lazy val contentSearchModeFormat: Format[ContentSearchMode] = {
+    implicitly[Format[String]].inmap(ContentSearchMode.fromString, ContentSearchMode.toString)
+  }
+
+  def fromString(s: String): ContentSearchMode = s match {
+    case "exists" => Exists
+  }
+
+  def toString(mode: ContentSearchMode): String = mode match {
+    case Exists => "exists"
+  }
+}
+
+final case class ProjectSearchRequest(filter: ProjectSearchFilter)
+object ProjectSearchRequest {
+  implicit lazy val projectSearchAdditionRequestFormat: OFormat[ProjectSearchRequest] =
+    Json.format[ProjectSearchRequest]
+}
+
+final case class ProjectSearchResponse(data: Seq[Project])
+object ProjectSearchResponse {
+  implicit lazy val projectSearchResponseFormat: OFormat[ProjectSearchResponse] =
+    Json.format[ProjectSearchResponse]
+}
--- a/services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
+++ b/services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
@@ -1,0 +1,142 @@
+package cromwell.pipeline.service
+
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.model.wrapper.UserId
+import cromwell.pipeline.service.ProjectConfigurationService.Exceptions.NotFound
+import cromwell.pipeline.service.ProjectSearchService.Exceptions._
+import cromwell.pipeline.service.exceptions.ServiceException
+
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success }
+
+trait ProjectSearchService {
+
+  def searchProjects(request: ProjectSearchRequest, userId: UserId): Future[ProjectSearchResponse]
+
+}
+
+object ProjectSearchService {
+
+  object Exceptions {
+    sealed abstract class ProjectSearchException(message: String) extends ServiceException(message)
+    final case class InternalError(message: String = "Internal error") extends ProjectSearchException(message)
+  }
+
+  // scalastyle:off method.length
+  def apply(
+    projectService: ProjectService,
+    projectFileService: ProjectFileService,
+    projectConfigurationService: ProjectConfigurationService
+  )(
+    implicit ec: ExecutionContext
+  ): ProjectSearchService =
+    new ProjectSearchService {
+
+      override def searchProjects(request: ProjectSearchRequest, userId: UserId): Future[ProjectSearchResponse] = {
+
+        def doSearch(searchFilter: ProjectSearchFilter, projects: Seq[Project]): Future[Seq[Project]] =
+          searchFilter match {
+            case All                   => Future.successful(projects)
+            case ByName(mode, value)   => Future.successful(searchByNameFilter(mode, value, projects))
+            case ByFiles(mode, value)  => searchByFiles(mode, value, projects, userId)
+            case ByConfig(mode, value) => searchByConfig(mode, value, projects, userId)
+
+            case Or(leftFilter, rightFilter) =>
+              for {
+                leftRes <- doSearch(leftFilter, projects)
+                rightRes <- doSearch(rightFilter, projects)
+              } yield (leftRes ++ rightRes).distinct
+
+            case And(leftFilter, rightFilter) =>
+              for {
+                leftRes <- doSearch(leftFilter, projects)
+                result <- doSearch(rightFilter, leftRes)
+              } yield result
+          }
+
+        for {
+          userProjects <- projectService.getUserProjects(userId).recoverWith {
+            case _ => internalError("fetch projects")
+          }
+          filteredProjects <- doSearch(request.filter, userProjects)
+        } yield ProjectSearchResponse(filteredProjects)
+
+      }
+
+      private def searchByConfig(
+        mode: ContentSearchMode,
+        value: Boolean,
+        projects: Seq[Project],
+        userId: UserId
+      ): Future[Seq[Project]] =
+        mode match {
+          case Exists => searchByConfigExists(value, projects, userId)
+        }
+
+      private def searchByConfigExists(
+        hasConfig: Boolean,
+        projects: Seq[Project],
+        userId: UserId
+      ): Future[Seq[Project]] = {
+        def getProjectWithConfig(project: Project): Future[(Project, Option[ProjectConfiguration])] =
+          projectConfigurationService.getLastByProjectId(project.projectId, userId).transformWith {
+            case Success(configuration) => Future.successful(project, Some(configuration))
+            case Failure(NotFound(_))   => Future.successful(project, None)
+            case _                      => internalError("fetch configuration")
+          }
+
+        for {
+          projWithConf <- Future.sequence(projects.map(getProjectWithConfig))
+        } yield projWithConf.collect {
+          case (proj, Some(_)) if hasConfig => proj
+          case (proj, None) if !hasConfig   => proj
+        }
+      }
+
+      private def searchByFiles(
+        mode: ContentSearchMode,
+        value: Boolean,
+        projects: Seq[Project],
+        userId: UserId
+      ): Future[Seq[Project]] =
+        mode match {
+          case Exists => searchByFilesExist(value, projects, userId)
+        }
+
+      private def searchByFilesExist(
+        hasFiles: Boolean,
+        projects: Seq[Project],
+        userId: UserId
+      ): Future[Seq[Project]] = {
+        def getProjectWithFiles(project: Project): Future[(Project, List[ProjectFile])] =
+          projectFileService
+            .getFiles(project.projectId, None, userId)
+            .recoverWith {
+              case _ => internalError("fetch files")
+            }
+            .map(files => (project, files))
+
+        for {
+          projWithFiles <- Future.sequence(projects.map(getProjectWithFiles))
+        } yield projWithFiles.collect {
+          case (proj, list) if list.nonEmpty == hasFiles => proj
+        }
+      }
+
+      private def searchByNameFilter(
+        mode: NameSearchMode,
+        value: String,
+        userProjects: Seq[Project]
+      ): Seq[Project] =
+        mode match {
+          case FullMatch   => userProjects.filter(_.name == value)
+          case RegexpMatch => userProjects.filter(_.name.matches(value))
+        }
+
+      private def internalError(action: String): Future[Nothing] =
+        Future.failed(InternalError(s"Failed to $action due to unexpected internal error"))
+
+    }
+  // scalastyle:on method.length
+
+}
--- a/services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
+++ b/services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
@@ -1,0 +1,208 @@
+package cromwell.pipeline.service
+
+import cromwell.pipeline.datastorage.dao.utils.TestProjectUtils
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.model.wrapper.UserId
+import cromwell.pipeline.service.ProjectSearchService.Exceptions.InternalError
+import cromwell.pipeline.service.impls.{
+  ProjectConfigurationServiceTestImpl,
+  ProjectFileServiceTestImpl,
+  ProjectServiceTestImpl
+}
+import org.scalatest.{ AsyncWordSpec, Matchers }
+
+import java.nio.file.Paths
+
+class ProjectSearchServiceTest extends AsyncWordSpec with Matchers {
+
+  private val dummyProject1: Project = TestProjectUtils.getDummyProject(name = "First")
+  private val projectId: ProjectId = dummyProject1.projectId
+  private val userId: UserId = dummyProject1.ownerId
+  private val dummyProject2 = TestProjectUtils.getDummyProject(ownerId = userId, name = "Second")
+  private val dummyProject3 = TestProjectUtils.getDummyProject(ownerId = userId, name = "Third")
+  private val dummyProjects = Seq(dummyProject1, dummyProject2, dummyProject3)
+
+  private val correctWdl = "task hello {}"
+  private val projectFilePath = Paths.get("test.txt")
+  private val projectFileContent = ProjectFileContent(correctWdl)
+  private val projectFile = ProjectFile(projectFilePath, projectFileContent)
+  private val dummyProject1FileBundle = (dummyProject1.projectId, Seq(projectFile))
+  private val dummyProject2FileBundle = (dummyProject2.projectId, Seq(projectFile))
+  private val dummyProject3FileBundle = (dummyProject3.projectId, Seq.empty)
+
+  private val wdlParams: WdlParams =
+    WdlParams(Paths.get("/home/file"), List(FileParameter("nodeName", StringTyped(Some("hello")))))
+  private val projectConfigurationId = ProjectConfigurationId.randomId
+
+  private val activeConfiguration1: ProjectConfiguration =
+    ProjectConfiguration(
+      projectConfigurationId,
+      projectId,
+      active = true,
+      wdlParams,
+      ProjectConfigurationVersion.defaultVersion
+    )
+  private val activeConfiguration2: ProjectConfiguration =
+    activeConfiguration1.copy(projectId = dummyProject2.projectId)
+
+  private val defaultProjectService = ProjectServiceTestImpl(dummyProjects: _*)
+  private val defaultProjectFileService =
+    ProjectFileServiceTestImpl(dummyProject1FileBundle, dummyProject2FileBundle, dummyProject3FileBundle)
+  private val defaultProjectConfigurationService =
+    ProjectConfigurationServiceTestImpl(activeConfiguration1, activeConfiguration2)
+
+  private def createProjectSearchService(
+    projectService: ProjectService = defaultProjectService,
+    projectFileService: ProjectFileService = defaultProjectFileService,
+    projectConfigurationService: ProjectConfigurationService = defaultProjectConfigurationService
+  ): ProjectSearchService =
+    ProjectSearchService(projectService, projectFileService, projectConfigurationService)
+
+  private val projectSearch = createProjectSearchService()
+
+  "ProjectSearchServiceTest" when {
+    "search successfully for all projects" should {
+      "return list of all projects" in {
+        val request = ProjectSearchRequest(All)
+        val result = ProjectSearchResponse(dummyProjects)
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+    }
+
+    "search project by name full match" should {
+      "return full matched project" in {
+        val name = dummyProject1.name
+        val filter = ByName(FullMatch, name)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject1))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "return empty result if nothing matched" in {
+        val projectName = "UNMATCHED"
+        val filter = ByName(RegexpMatch, projectName)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq.empty)
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+    }
+
+    "search project by name regexp match" should {
+      "return matched projects" in {
+        val regexString = ".*ir.*"
+        val filter = ByName(RegexpMatch, regexString)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject1, dummyProject3))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "return empty result if nothing matched" in {
+        val regexString = ".*UNMATCHED.*"
+        val filter = ByName(RegexpMatch, regexString)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq.empty)
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+    }
+
+    "search project by config" should {
+      "return projects with configuration if Exists true" in {
+        val filter = ByConfig(mode = Exists, value = true)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject1, dummyProject2))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "return project without configuration if Exists false" in {
+        val filter = ByConfig(mode = Exists, value = false)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject3))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "fail with exception if unable to fetch configuration" in {
+        val failedConfigService = ProjectConfigurationServiceTestImpl.withException(new RuntimeException)
+        val projectSearch = createProjectSearchService(projectConfigurationService = failedConfigService)
+
+        val filter = ByConfig(mode = Exists, value = true)
+        val request = ProjectSearchRequest(filter)
+
+        projectSearch.searchProjects(request, userId).failed.map {
+          _ shouldBe InternalError("Failed to fetch configuration due to unexpected internal error")
+        }
+      }
+    }
+
+    "search project by files" should {
+      "return projects with files if Exists true" in {
+        val filter = ByFiles(mode = Exists, value = true)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject1, dummyProject2))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "return project without files if Exists false" in {
+        val filter = ByFiles(mode = Exists, value = false)
+        val request = ProjectSearchRequest(filter)
+        val result = ProjectSearchResponse(Seq(dummyProject3))
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe result)
+      }
+      "fail with exception if unable to fetch configuration" in {
+        val failedFilesService = ProjectFileServiceTestImpl.withException(new RuntimeException)
+        val projectSearch = createProjectSearchService(projectFileService = failedFilesService)
+
+        val filter = ByFiles(mode = Exists, value = true)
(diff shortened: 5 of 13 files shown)


services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
```diff
@@ -127,5 +127,6 @@ object ProjectService {
 
       private def internalError(action: String): Future[Nothing] =
         Future.failed(InternalError(s"Failed to $action due to unexpected internal error"))
+
```
gominasian: Can't get the point of this newline. Is it some formatter thing?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+      private def getProjectsByFullNameMatch(name: String, userProjects: Future[Seq[Project]]): Future[Seq[Project]] =
+        userProjects.map(projects => projects.filter(_.name == name))
+
+      private def getProjectsByRegexpMatch(
+        nameRegexp: String,
+        userProjects: Future[Seq[Project]]
+      ): Future[Seq[Project]] =
+        userProjects.map(projects => projects.filter(_.name.matches(nameRegexp)))
```
DruzhininPavel: Do you really need these two private methods? Do you want to reuse them ?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,110 @@
...
+
+          for {
+            projects <- eventualProjects
+            projWithFiles <- Future.sequence(projects.map(getProjectWithFiles))
+          } yield projWithFiles.collect {
+            case (proj, list) if list.nonEmpty == hasFiles => Some(proj)
+            case _                                         => None
+          }.flatten
+
+        case _ => Future.failed(new RuntimeException("No such search type defined"))
```
DruzhininPavel: Do we really need this case with Future.filed? maybe we don't need to use "match"?

DruzhininPavel (changes requested):

services/src/main/scala/cromwell/pipeline/service/ProjectService.scala
Nikolai-Isaev: Oops, fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: That was an artifact from the past, thanks

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
DruzhininPavel: Ok.

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
DruzhininPavel: Yeap. Looks good.

DruzhininPavel (approved):
LGTM

gominasian (approved):
We may need postman collection for this. Everything else is fine for me

Nikolai-Isaev:
Thanks, Postman collection added

controllers/src/main/scala/cromwell/pipeline/controller/ControllerModule.scala
```diff
@@ -18,4 +18,7 @@ class ControllerModule(serviceModule: ServiceModule, authConfig: AuthConfig)(
   lazy val projectFileController: ProjectFileController = new ProjectFileController(serviceModule.projectFileService)
   lazy val configurationController = new ProjectConfigurationController(serviceModule.configurationService)
   lazy val runController: RunController = new RunController(serviceModule.runService)
+  lazy val projectSearchController: ProjectSearchController = new ProjectSearchController(
+    serviceModule.projectSearchService
+  )
```
myazinn: [minor] can we make it prettier please?
```suggestion
  lazy val projectSearchController: ProjectSearchController =
    new ProjectSearchController(serviceModule.projectSearchService)
```

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  private def searchProjects(implicit accessToken: AccessTokenContent): Route = post {
+    entity(as[ProjectSearchRequest]) { request =>
+      complete(projectSearchService.searchProjects(request, accessToken.userId))
+    }
+  }
+
+  val route: AccessTokenContent => Route = implicit accessToken =>
+    handleExceptions(projectSearchExceptionHandler) {
+      pathPrefix("search") {
```
myazinn: Ticket description says it should be `projects/search`, not just `search`

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
```diff
@@ -0,0 +1,33 @@
...
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.service.ProjectSearchService
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+
+class ProjectSearchController(projectSearchService: ProjectSearchService) {
+
+  private def searchProjects(implicit accessToken: AccessTokenContent): Route = post {
+    entity(as[ProjectSearchRequest]) { request =>
+      complete(projectSearchService.searchProjects(request, accessToken.userId))
```
myazinn: The scheme of the returned object here does not correspond to what's required by ticket. It should be 
```
{
  data: ["here go projects"]
}
```
and not
```json
["here go projects"]
```
This is required to make API extensible and be able to add more metainformation in the future (like filter id or pagination information)

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
```diff
@@ -0,0 +1,33 @@
...
+import cromwell.pipeline.controller.ProjectSearchController.projectSearchExceptionHandler
+import cromwell.pipeline.datastorage.dto._
+import cromwell.pipeline.datastorage.dto.auth.AccessTokenContent
+import cromwell.pipeline.service.ProjectSearchService
+import de.heikoseeberger.akkahttpplayjson.PlayJsonSupport._
+
+class ProjectSearchController(projectSearchService: ProjectSearchService) {
+
+  private def searchProjects(implicit accessToken: AccessTokenContent): Route = post {
+    entity(as[ProjectSearchRequest]) { request =>
```
myazinn: You also need to add `pathEnd` here. Otherwise it will work with unwanted paths like `search/foo/bar/buzz` which might lead to conflicts

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,68 @@
...
+sealed trait ProjectSearchFilter
+object ProjectSearchFilter {
+  implicit lazy val projectSearchFilterReads: Reads[ProjectSearchFilter] = Reads[ProjectSearchFilter] { json =>
+    (json \ "type").validate[String].flatMap {
+      case "all"            => json.validate[AllFilter.type]
+      case "name"           => json.validate[ByNameFilter]
+      case "files"          => json.validate[ByFilesFilter]
+      case "configurations" => json.validate[ByConfigFilter]
+      case "OR"             => json.validate[ComplexOrFilter]
+      case "AND"            => json.validate[ComplexAndFilter]
```
myazinn: it should be `or` and `and` (i.e. lowercased), isn't it?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,68 @@
+package cromwell.pipeline.datastorage.dto
+
+import play.api.libs.json._
+
+sealed trait ProjectSearchFilter
+object ProjectSearchFilter {
+  implicit lazy val projectSearchFilterReads: Reads[ProjectSearchFilter] = Reads[ProjectSearchFilter] { json =>
+    (json \ "type").validate[String].flatMap {
+      case "all"            => json.validate[AllFilter.type]
+      case "name"           => json.validate[ByNameFilter]
```
myazinn: I think we can be better then this.
Play Json has support for custom naming strategy and using discriminator field which can help with automatic deserilization of sealed trait.
https://www.playframework.com/documentation/2.8.x/ScalaJsonAutomated
Also, take a look at `cromwell.pipeline.utils.json.AdtJsonFormatter` and its usages in `cromwell.pipeline.datastorage.dto.TypedValue#typedValueFormat`. This solves the same issue in a different (maybe more handy in this case) way.
So, please use existing tools.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,68 @@
...
+      case "files"          => json.validate[ByFilesFilter]
+      case "configurations" => json.validate[ByConfigFilter]
+      case "OR"             => json.validate[ComplexOrFilter]
+      case "AND"            => json.validate[ComplexAndFilter]
+    }
+  }
+}
+
+case object AllFilter extends ProjectSearchFilter {
```
myazinn: [minor] I think we can drop `Filter` part everywhere to be less verbose. I.e.
`case object All extends ProjectSearchFilter`
`final case class ByName(mode: NameSearchMode, value: String) extends ProjectSearchFilter`
etc

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,68 @@
...
+    implicitly[Reads[String]].map(NameSearchMode.fromString)
+  }
+
+  def fromString(s: String): NameSearchMode = s match {
+    case "full_match"   => FullMatch
+    case "regexp_match" => RegexpMatch
+  }
+}
+
+final case class ByConfigFilter(mode: String, value: Boolean) extends ProjectSearchFilter
```
myazinn: I don't see any usages of `mode` field anywhere (same for `ByFilesFilter`). If it's not required, why do we need it? If it's required, why it's unused?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,68 @@
...
+object ByConfigFilter {
+  implicit lazy val byConfigFilterReads: Reads[ByConfigFilter] = Json.reads[ByConfigFilter]
+}
+
+final case class ByFilesFilter(mode: String, value: Boolean) extends ProjectSearchFilter
+object ByFilesFilter {
+  implicit lazy val byFilesFilterReads: Reads[ByFilesFilter] = Json.reads[ByFilesFilter]
+}
+
+final case class ComplexOrFilter(left: ProjectSearchFilter, right: ProjectSearchFilter) extends ProjectSearchFilter
```
myazinn: Can't we just name them as `Or` and `And` instead of `ComplexOrFilter` and `ComplexAndFilter` ?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,115 @@
...
+
+        def doSearch(searchFilter: ProjectSearchFilter): Future[Seq[Project]] = searchFilter match {
+          case AllFilter                    => userProjects
+          case ByNameFilter(mode, value)    => searchWithNameFilter(mode, value, userProjects)
+          case ByFilesFilter(_, hasFiles)   => searchByFilesFilter(hasFiles, userProjects, userId)
+          case ByConfigFilter(_, hasConfig) => searchByConfigFilter(hasConfig, userProjects, userId)
+          case ComplexOrFilter(leftFilter, rightFilter) =>
+            doSearch(leftFilter).flatMap(
+              resultLeft => doSearch(rightFilter).map(resultRight => resultLeft ++ resultRight)
+            )
```
myazinn: Shouldn't we make sure here that the result contains only unique values?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,115 @@
...
+          case ComplexAndFilter(leftFilter, rightFilter) =>
+            doSearch(leftFilter).flatMap(
+              resultLeft => doSearch(rightFilter).map(resultRight => resultLeft.intersect(resultRight))
+            )
```
myazinn: Why do we need to traverse all user projects with left filter, then traverse _the same_ projects with right filter and then search for intersection? Can't we feed to right filter the output of the left filter? That will give us their intersection for free and might be much more efficient

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,115 @@
...
+        }
+        doSearch(request.filter).map(_.distinct)
+      }
+
+      private def searchByConfigFilter(
+        hasConfig: Boolean,
+        eventualProjects: Future[Seq[Project]],
```
myazinn: Methods accepting a `Future` are often signatures that the code "smells". Do these methods really need a `Future[Seq[Project]]` as an input? What's wrong with just `projects: Seq[Project]`?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,115 @@
...
+          }
+
+        for {
+          projects <- eventualProjects
+          projWithConf <- Future.sequence(projects.map(getProjectWithConfig))
+        } yield projWithConf.collect {
+          case (proj, Some(_)) if hasConfig => Some(proj)
+          case (proj, None) if !hasConfig   => Some(proj)
+          case _                            => None
+        }.flatten
```
myazinn: `case _ =>` in `collect` function usually indicates that you are doing something wrong. It doesn't make sense to collect data with total function, with total functions we use `map` / `flatMap`
```suggestion
        } yield projWithConf.collect {
          case (proj, Some(_)) if hasConfig => proj
          case (proj, None) if !hasConfig   => proj
        }
```

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,115 @@
...
+              case _ => internalError("fetch files")
+            }
+            .map(files => (project, files))
+
+        for {
+          projects <- eventualProjects
+          projWithFiles <- Future.sequence(projects.map(getProjectWithFiles))
+        } yield projWithFiles.collect {
+          case (proj, list) if list.nonEmpty == hasFiles => Some(proj)
+          case _                                         => None
```
myazinn: Same comment about `collect` here

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
```diff
@@ -0,0 +1,104 @@
...
+  ProjectConfigurationServiceTestImpl,
+  ProjectFileServiceTestImpl,
+  ProjectServiceTestImpl
+}
+import org.scalatest.{ AsyncWordSpec, Matchers }
+import org.scalatestplus.mockito.MockitoSugar
+
+import java.nio.file.Paths
+
+class ProjectSearchServiceTest extends AsyncWordSpec with Matchers with MockitoSugar {
```
myazinn: What about API (`ProjectSearchController`) tests?

myazinn: Why do we need `MockitoSugar`  here?

services/src/test/scala/cromwell/pipeline/service/impls/ProjectConfigurationServiceTestImpl.scala
```diff
@@ -19,7 +19,12 @@ class ProjectConfigurationServiceTestImpl(projectConfigurations: Seq[ProjectConf
   override def getLastByProjectId(projectId: ProjectId, userId: UserId): Future[ProjectConfiguration] =
     testMode match {
       case WithException(exc) => Future.failed(exc)
-      case _                  => Future.successful(projectConfigurations.head)
+
+      case _ =>
+        projectConfigurations.find(conf => conf.projectId == projectId && conf.active) match {
```
myazinn: That looks like a weird ad-hoc solution. Why do you need to filter by configuration activity? Can't you just not pass inactive configurations to this test implementation if you don't wan't them to appear in the result?

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
```diff
@@ -0,0 +1,104 @@
...
+    "search project with complex filter" should {
+      "return projects without config or name match" in {
+        val name = dummyProject1.name
+        val filterByName = ByNameFilter(FullMatch, name)
+        val filterByConfig = ByConfigFilter("exists", value = false)
+        val complexFilter = ComplexOrFilter(filterByName, filterByConfig)
+        val request = ProjectSearchRequest(complexFilter)
+
+        projectSearch.searchProjects(request, userId).map(_ shouldBe Seq(dummyProject1, dummyProject3))
+      }
```
myazinn, line 205: I'm seeing 6 different filter types and only 5 tests (2 of them are for `ByNameFilter`)  which means some filters are not covered by tests. Can you please the missing ones?
Also, filters like `ByConfigFilter` and `ByFileFilter` accept boolean parameter which means we should have at least two tests for each.

myazinn (changes requested):

controllers/src/main/scala/cromwell/pipeline/controller/ControllerModule.scala
Nikolai-Isaev: Done

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
Nikolai-Isaev: Fixed

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
Nikolai-Isaev: Fixed

controllers/src/main/scala/cromwell/pipeline/controller/ProjectSearchController.scala
Nikolai-Isaev: Fixed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Removed

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: Added

Nikolai-Isaev: Eradicated

services/src/test/scala/cromwell/pipeline/service/impls/ProjectConfigurationServiceTestImpl.scala
Nikolai-Isaev: Fixed

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: There is a problem with current byFile search testing. Since files and projects are not connected by id it is impossible to filter files by project with current testImpl. So I tried to temporary solve that by making a new `ProjectFileServiceIdCombinedTestImpl` (masterful naming skills applied) which accepts  `(ProjectId, Seq[ProjectFile])` tuple. With that we can receive different search results for different projects during testing.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Fixed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: `Distinct` added

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Refactored

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Refactored

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Removed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: As for now I believe we can replace {"mode" : "exists", "value" : boolean} with {"exists" : boolean} 
Maybe the idea was to add more modes for config and file search?
P.S. adt for lonely Exist mode created

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,101 @@
...
+import play.api.libs.functional.syntax.toInvariantFunctorOps
+import play.api.libs.json.JsonConfiguration.Aux
+import play.api.libs.json._
+
+sealed trait ProjectSearchFilter
+object ProjectSearchFilter {
+  implicit lazy val jsonConfig: Aux[Json.MacroOptions] = JsonConfiguration(
+    discriminator = "type",
+    typeNaming = JsonNaming { fullName =>
+      fullName.split("\\.").lastOption.fold("")(_.toLowerCase)
```
myazinn: Does it meet ticket requirements? I guess it will give us something like this `"type": "ByName"`, while ticket requires it to be `"type": "name"`. Based on postman collection I see a discrepancy as well

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,114 @@
...
+    implicit ec: ExecutionContext
+  ): ProjectSearchService =
+    new ProjectSearchService {
+
+      override def searchProjects(request: ProjectSearchRequest, userId: UserId): Future[ProjectSearchResponse] = {
+        val userProjects = projectService.getUserProjects(userId).recoverWith {
+          case _ => internalError("fetch projects")
+        }
+
+        def doSearch(searchFilter: ProjectSearchFilter, projects: Future[Seq[Project]]): Future[Seq[Project]] =
```
myazinn: Why does it still accept `Future[Seq[Project]]`? Why can't it just accept `Seq[Project]` ?

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,101 @@
...
+    case "regexp_match" => RegexpMatch
+  }
+
+  def toString(mode: NameSearchMode): String = mode match {
+    case FullMatch   => "full_match"
+    case RegexpMatch => "regexp_match"
+  }
+}
+
+final case class ByConfig(mode: ContentSearchMode, value: Boolean) extends ProjectSearchFilter
```
myazinn: I still don't see any usages of `mode` field

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,114 @@
...
+            case Failure(NotFound(_))   => Future.successful(project, None)
+            case _                      => internalError("fetch configuration")
+          }
+
+        for {
+          projWithConf <- Future.sequence(projects.map(getProjectWithConfig))
+        } yield projWithConf.collect {
+          case (proj, Some(_)) if hasConfig => Some(proj)
+          case (proj, None) if !hasConfig   => Some(proj)
+        }.flatten
```
myazinn: Why do you need to wrap all possible cases to `Some(_)` and then flatten it? Can't you just take result as is?
```suggestion
        } yield projWithConf.collect {
          case (proj, Some(_)) if hasConfig => proj
          case (proj, None) if !hasConfig   => proj
        }
```
Please have a look at documentation on how to work with `collect` method. Same applies to `searchByFilesFilter`

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
```diff
@@ -0,0 +1,73 @@
...
+    path: Path,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[ProjectFile] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _ =>
+        val filesOpt = projectFileBundles.find {
+          case (id, _) => id == projectId
+        }.map(_._2)
```
myazinn: `collectFirst`

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
```diff
@@ -0,0 +1,196 @@
...
+        val filter = ByFiles(mode = Exists, value = true)
+        val request = ProjectSearchRequest(filter)
+
+        projectSearch.searchProjects(request, userId).failed.map {
+          _ shouldBe InternalError("Failed to fetch files due to unexpected internal error")
+        }
+      }
+    }
+
+    "search project with complex filter" should {
```
myazinn, line 179: I think it would be useful to add tests on `And` filter that shows that projects that satisfy only some of the conditions won't be included to the result. The only existing test doesn't show that `And` filter can actually exclude anything.

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
myazinn: I think it's fine to have a way to specify project id for project file, but I don't think it's a good idea to have several ways to create test implementation. You should create an overloaded version of `ProjectFileServiceTestImpl.apply` method that accepts `(ProjectId, ProjectFile)` instead of `ProjectFile`. Under the hood it might create another implementation, but it should still be single point of test implementation creation. And yeah, the name can be better I think.

myazinn: Or maybe we can modify existing `ProjectFileServiceTestImpl` to accept `(ProjectId, ProjectFile)` instead of `ProjectFile`, I think that is a valid choice too.

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
```diff
@@ -0,0 +1,73 @@
...
+        val file = filesOpt.flatMap(seq => seq.find(_.path == path)).get
```
myazinn: Why not return `NotFound` in this case?

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
```diff
@@ -0,0 +1,73 @@
...
+    projectId: ProjectId,
+    version: Option[PipelineVersion],
+    userId: UserId
+  ): Future[List[ProjectFile]] =
+    testMode match {
+      case WithException(exc) => Future.failed(exc)
+      case _ =>
+        val files = projectFileBundles.find {
+          case (id, _) => id == projectId
+        }.map(_._2).get.toList
```
myazinn: What if it's not found? Aren't you supposed to just return an empty list? Do you really need to call `get` on `Option` in this case?

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
myazinn: > Maybe the idea was to add more modes for config and file search?
The idea was to make addition of more modes possible in the future. So even though now it is the only mode, we might want to add more in any moment

myazinn: > Maybe the idea was to add more modes for config and file search?

The idea was to make addition of more modes possible in the future. So even though now it is the only mode, we might want to add more in any moment. What happens if tomorrow we decide to add something like this?
```
{
  "type": "files",
  "mode: "contain",
  "value": "some_file_name.txt"
}
```

Will addition of new adt case `Contain` cover it?  Does your current case class structure allow such changes now? If I add this case, will the compiler notice it and warn me / show an error, that there is no handling of this case in ProjectSearchService?

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
myazinn: Your changes do not contain everything that was suggested here

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: `ProjectFileServiceTestImpl` modified to accept `(ProjectId, ProjectFile)`

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Refactored

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Fixed

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
Nikolai-Isaev: Refactored

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: Test added

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
Nikolai-Isaev: Method recreated

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceIdCombinedTestImpl.scala
Nikolai-Isaev: Method recreated

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: That is a major flaw in current implementation. For now addition of new mode will work fine only with specific type of `value`. It should be rethought and rebuilt

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: `ProjectFileServiceTestImpl` modified to accept `(ProjectId, ProjectFile)`

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Fixed. It works but total refactoring needed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,103 @@
...
+        case "byfiles"  => "files"
+        case _          => jsName
+      }
+    }
+  )
+  implicit lazy val projectSearchFilterFormat: OFormat[ProjectSearchFilter] = Json.format[ProjectSearchFilter]
+}
+
+case object All extends ProjectSearchFilter {
+  implicit lazy val allFormat: OFormat[All.type] = AdtJsonFormatter.objectFormat(All)
```
myazinn: Since you've already found this class (`AdtJsonFormatter.objectFormat`), why not use it for `ProjectSearchFilter` and not only for one of its cases? I imagine it will make code cleaner

scalastyle-config.xml
```diff
@@ -64,7 +64,7 @@
  </check>
  <check level="warning" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxLength"><![CDATA[100]]></parameter>
+   <parameter name="maxLength"><![CDATA[150]]></parameter>
```
myazinn: Can you please make a ticket in Jira to fix this issue?

myazinn: Also, I think it's better to disable scalastyle in one particular place instead of changing it globally

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
myazinn: I don't see this test. Right now your implementation has a bug. I expect you to write a test that can catch it and then of course fix this bug.

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceTestImpl.scala
```diff
@@ -7,7 +7,8 @@ import cromwell.pipeline.service.{ ProjectFileService, VersioningException }
 import java.nio.file.Path
 import scala.concurrent.Future
 
-class ProjectFileServiceTestImpl(projectFiles: Seq[ProjectFile], testMode: TestMode) extends ProjectFileService {
+class ProjectFileServiceTestImpl(projectFileBundles: Seq[(ProjectId, Seq[ProjectFile])], testMode: TestMode)
```
myazinn: Sorry for not saying about it earlier, but you should definitely use `Map[ProjectId, Seq[ProjectFile]]` here

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceTestImpl.scala
```diff
@@ -46,17 +54,23 @@ class ProjectFileServiceTestImpl(projectFiles: Seq[ProjectFile], testMode: TestM
...
     testMode match {
       case WithException(exc) => Future.failed(exc)
-      case _                  => Future.successful(projectFiles.toList)
+      case _ =>
+        val filesOpt = projectFileBundles.collectFirst {
+          case (id, files) if id == projectId => files
+        }
+        filesOpt match {
+          case Some(files) => Future.successful(files.toList)
+          case None        => Future.failed(new RuntimeException("No correct test file bundle received"))
```
myazinn: Why not just return an empty list? It sounds reasonable

myazinn (changes requested):

scalastyle-config.xml
GrigoriiBerezin: I thought we talked about it and the decision was to disable scalastyle in one file

myazinn: Yep, that was my thoughts as well

services/src/test/scala/cromwell/pipeline/service/ProjectSearchServiceTest.scala
Nikolai-Isaev: That was an embarrassing one. In `projectSearchService` initial Seq of projects was always used inside `doSearch` method.
Test added, `doSearch` fixed

scalastyle-config.xml
Nikolai-Isaev: Instead of global allowed method length change, apply method in ProjectSearchService was wrapped in "scalastyle method length off" comment

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: Fixed, looks prettier and more simple indeed

services/src/test/scala/cromwell/pipeline/service/impls/ProjectFileServiceTestImpl.scala
Nikolai-Isaev: Fixed

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
```diff
@@ -0,0 +1,141 @@
...
+    projectConfigurationService: ProjectConfigurationService
+  )(
+    implicit ec: ExecutionContext
+  ): ProjectSearchService =
+    new ProjectSearchService {
+
+      override def searchProjects(request: ProjectSearchRequest, userId: UserId): Future[ProjectSearchResponse] = {
+        val eventualUserProjects = projectService.getUserProjects(userId).recoverWith {
+          case _ => internalError("fetch projects")
+        }
```
myazinn: [minor] do you still need it here? It only causes confusion as you've already seen. Can't you just inline this variable at the only usage?

myazinn (approved):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
myazinn: Hmm, sorry, I've approved MR thinking that there is only one minor thing, but I see that this comment is not resolved yet

myazinn (changes requested):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Refactored with `AdtJsonFormatter`

services/src/main/scala/cromwell/pipeline/service/ProjectSearchService.scala
Nikolai-Isaev: Removed

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Made a private method to generalize formatter creation for `Or` and `And` filters. Couldn't make `projectSearchFilter: OFormat` to be parameter though.

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
```diff
@@ -0,0 +1,92 @@
...
+final case class Or(left: ProjectSearchFilter, right: ProjectSearchFilter) extends ProjectSearchFilter
+final case class And(left: ProjectSearchFilter, right: ProjectSearchFilter) extends ProjectSearchFilter
+
+object ProjectSearchFilter {
+  implicit val allFormat: OFormat[All.type] = objectFormat(All)
+  implicit val byNameFormat: OFormat[ByName] = Json.format
+  implicit val byConfigFormat: OFormat[ByConfig] = Json.format
+  implicit val byFilesFormat: OFormat[ByFiles] = Json.format
+  implicit val orFormat: OFormat[Or] = getComplexFilterFormat(Or, Or.unapply)
+  implicit val andFormat: OFormat[And] = getComplexFilterFormat(And, And.unapply)
```
myazinn: [minor]
One last thing, can we please either mark them private or hide them inside
```scala
implicit val projectSearchFilterFormat: OFormat[ProjectSearchFilter] = {
  implicit val allFormat: OFormat[All.type] = objectFormat(All)
  // etc
}
```
This is required to hide these formatters and not let others to accidentally try to serialize / deserialize filters without type descriptor.

myazinn (approved):

repositories/src/main/scala/cromwell/pipeline/datastorage/dto/ProjectSearch.scala
Nikolai-Isaev: Done

GrigoriiBerezin (approved):
Great job

PR #236, opened 2022-02-07
EPMLSTRCMW-245 Test reports are not generated for some classes

ForeverRainmaN:
Reconfigured build.sbt with proper flags passed into testOptions:
-o means standard output, -h saves reports as xml files inside specific directory, that's been passed as the last argument


GrigoriiBerezin (approved):
Clear

myazinn (approved):

